PR #4, opened 2025-06-16
Root chain integration

RaimoMeltsas:


bft-config/keys.json
```diff
@@ -0,0 +1,10 @@
+{
+  "sigKey": {
+    "algorithm": "secp256k1",
+    "privateKey": "0xad6acf4eb44dc87912c075824a49f5cfb468224bbd6aae0cabd3aa5286b42c34"
```
MastaP: Let's cleanup these files or at least rename them to *.json.example to avoid someone accidentally deploying them

internal/bft/client.go
```diff
@@ -0,0 +1,286 @@
...
+		roundManager: roundManager,
+	}
+}
+
+func (n *BFTClientStub) Start(ctx context.Context, nextRoundNumber *api.BigInt) error {
+	return n.roundManager.StartNewRound(ctx, nextRoundNumber)
+}
+
+func (n *BFTClientStub) CertificationRequest(ctx context.Context, block *models.Block) error {
+	n.roundManager.FinalizeBlock(ctx, block)
```
MastaP, line 281: FinalizeBlock is called twice, here and upon receiving cert response. is this correct?

bft-config/keys.json
RaimoMeltsas: Removed bft-config from repository. Added it also into gitignore.

internal/bft/client.go
RaimoMeltsas: This call is part of BFTClientStub, that is used in case BFT support is disabled. Moved BFTClientStub into separate go file for less confusion.

PR #16, opened 2025-07-04
Store block record when block is finalized

RaimoMeltsas:


internal/round/batch_processor.go
```diff
@@ -185,6 +185,21 @@ func (rm *RoundManager) FinalizeBlock(ctx context.Context, block *models.Block)
 		return fmt.Errorf("failed to store block: %w", err)
 	}
 
+	var requestIds []api.RequestID
+	for _, commitment := range rm.currentRound.Commitments {
+		requestIds = append(requestIds, commitment.RequestID)
+	}
+
+	blockRecord := &models.BlockRecords{
```
MastaP, line 193: there's a constructor NewBlockRecords()

internal/models/block.go
```diff
@@ -119,6 +119,27 @@ type BlockRecords struct {
 	CreatedAt   *api.Timestamp  `json:"createdAt" bson:"createdAt"`
 }
 
+type BlockRecordsBSON struct {
```
MastaP, line 122: BlockRecords already has bson tags on the fields. maybe implement MarshalBSON() methods on custom types instead of having BlockRecordsBSON struct? ToBSON() adds additional linear pass on the array of requests which could be potentially avoided

internal/round/batch_processor.go
RaimoMeltsas: Using NewBlockRecords.

internal/models/block.go
RaimoMeltsas: Removed BlockRecordsBSON and ToBSON(). Added Marshal/Unmarshal BSON for needed types.

PR #34, opened 2025-09-01
use smt snapshot for for block creation

MastaP:


internal/round/thread_safe_smt.go
```diff
@@ -109,3 +118,98 @@ func (ts *ThreadSafeSMT) WithWriteLock(fn func() error) error {
...
+		"leafCount":  0, // Could be enhanced to show actual count
+		"isSnapshot": true,
+		"isLocked":   false, // Could be enhanced to show lock status
+	}
+}
+
+// Commit commits the snapshot changes back to the original ThreadSafeSMT
+// This operation requires write locks on both the snapshot and the original SMT
+// to ensure atomicity of the commit operation
+func (tss *ThreadSafeSmtSnapshot) Commit(originalSMT *ThreadSafeSMT) error {
```
jait91, line 196: always returns nil error so error should be removed

PR #40, opened 2025-09-10
Fix invalid test

ahtotruu:


internal/smt/smt_test.go
```diff
@@ -1236,6 +1252,5 @@ func TestSMTSnapshot(t *testing.T) {
 	t.Logf("Order [1,3]: %s", hash1)
 	t.Logf("Order [3,1]: %s", hash2)
```
MastaP, line 1253: this also needs to change

MastaP (approved):

PR #51, opened 2025-09-20
Streaming round processing with adaptive timing

MastaP:
1. **Adaptive Processing Deadline**
   - Calculated as `roundDuration * processingRatio`
   - Starts at 90% (0.9) but adjusts based on performance

2. **Performance Tracking**
   - **SMT Update Time**: Tracks time spent updating the Sparse Merkle Tree
   - **Finalization Time**: Measures block proposal and BFT consensus time
   - **Processing Time**: Monitors actual commitment processing duration
   - Uses exponential moving averages (80/20 weight) for smooth adaptation

3. **Dynamic Ratio Adjustment** (`adjustProcessingRatio`)
   - Calculates ideal processing ratio based on average finalization time
   - Leaves 50ms safety buffer for finalization
   - Gradually adjusts ratio (80/20 weighted average)
   - Bounded between 50% and 95% for safety
   - Adapts to system performance changes

4. **Enhanced Metrics API**
   - New `adaptiveTiming` section in `GetStreamingMetrics()`
   - Shows current processing/finalization windows
   - Displays average timings for monitoring
   - Helps operators understand system behavior

### DIFF
--- a/cmd/performance-test/main.go
+++ b/cmd/performance-test/main.go
@@ -74,7 +74,7 @@ const (
 	defaultAggregatorURL = "http://localhost:3000"
 	testDuration         = 30 * time.Second
 	workerCount          = 30   // Number of concurrent workers
-	requestsPerSec       = 1000 // Target requests per second
+	requestsPerSec       = 5000 // Target requests per second
 )
 
 // BlockCommitmentInfo stores block number and commitment count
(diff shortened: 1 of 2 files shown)


b3y0urs3lf:
Tested locally using GO aggregator commit ac20262f97769190c06b9f69512ac171e3bc6485 and Java SDK commit and Java SDK commit 667a684f53c5ef304a7039cf532b781a6bd112f1.

1. Not in HA mode (1 aggregator running), although I submitted only 1 commit to GO aggregator, but I l see that commit count is 168 , from logs perspective there are a lot of "Resetting cursor to fetch new commitments" 
image
image
Expected result: when not in HA mode, counter should still work correctly. 
2. In HA mode, round processing with adaptive timing is not working at all, meaning above described issue is not reproduced.  But there is no point of moving this task to done as developed functionality in all modes (as in production we should use both HA mode and sharding etc.)

For now moving task back to-do.

PR #63, opened 2025-10-07
Redis commitment queue

jait91:


internal/storage/redis/commitment.go
```diff
@@ -0,0 +1,562 @@
...
+			return
+		}
+	}
+}
+
+// trimStream keeps the stream at a reasonable size
+func (cs *CommitmentStorage) trimStream(ctx context.Context) {
+	// Trim to keep only the last N messages
+	// Errors are silently ignored - this is best-effort maintenance
+	_ = cs.client.XTrimMaxLen(ctx, commitmentStream, maxStreamLength).Err()
```
lploom: Should probably somehow log the trimmed commitments?

internal/round/round_manager.go
```diff
@@ -13,6 +13,7 @@ import (
 	"github.com/unicitynetwork/aggregator-go/internal/models"
 	"github.com/unicitynetwork/aggregator-go/internal/smt"
 	"github.com/unicitynetwork/aggregator-go/internal/storage/interfaces"
+	"github.com/unicitynetwork/aggregator-go/internal/storage/redis"
```
lploom, line 16: ideally we should not depend on redis package in round manager

lploom (approved):

PR #72, opened 2025-10-22
Sharding support

jait91:
Add optional parent and child modes for aggregator for sharding


internal/models/authenticator.go
```diff
@@ -0,0 +1,59 @@
+package models
+
+import (
+	"fmt"
+
+	"github.com/unicitynetwork/aggregator-go/pkg/api"
+)
+
+// Authenticator represents the authentication data for a commitment
+type Authenticator struct {
```
MastaP: why this (third) struct is needed? Seems api.Authenticator and AuthenticatorBSON to be enough

internal/models/block.go
```diff
@@ -85,51 +91,50 @@ func (bb *BlockBSON) FromBSON() (*Block, error) {
...
+		NoDeletionProofHash:  noDeletionProofHash,
+		CreatedAt:            api.NewTimestamp(bb.CreatedAt),
+		UnicityCertificate:   unicityCertificate,
+		ParentMerkleTreePath: parentMerkleTreePath,
+	}, nil
 }
 
 // NewBlock creates a new block
-func NewBlock(index *api.BigInt, chainID, version, forkID string, rootHash, previousBlockHash api.HexBytes) *Block {
+func NewBlock(index *api.BigInt, chainID string, shardID int, version, forkID string, rootHash, previousBlockHash, uc api.HexBytes, parentMerkleTreePath *api.MerkleTreePath) *Block {
```
MastaP: use `shardID api.ShardID` instead of `int`

internal/round/parent_round_manager.go
```diff
@@ -0,0 +1,503 @@
...
+	if prm.currentRound == nil {
+		prm.roundMutex.Unlock()
+		return fmt.Errorf("no current round to process")
+	}
+
+	// Capture current round for processing
+	round := prm.currentRound
+	round.State = RoundStateProcessing
+
+	round.ProcessedShardUpdates = make(map[int]*models.ShardRootUpdate, len(round.ShardUpdates))
```
MastaP: since ShardUpdates is carried over from the previous round this always triggers the full tree update

internal/models/authenticator.go
lploom: We didn't add it, it was just moved to separate file. Looks like design choice to separate api and domain structs. Since it already exists, even though all the fields are the same, it's not worth refactoring IMO.

internal/models/block.go
lploom: Fine. But in general I don't understand the use of api types everywhere. What if you want to change ShardID to ShardIDV2? It would be absolute nightmare to do the migration.

internal/service/parent_service.go
```diff
@@ -0,0 +1,273 @@
...
+
+	if err := pas.validateShardID(req.ShardID); err != nil {
+		pas.logger.WithContext(ctx).Warn("Invalid shard ID", "shardId", req.ShardID, "error", err.Error())
+		return nil, fmt.Errorf("invalid shard ID: %w", err)
+	}
+
+	pas.logger.WithContext(ctx).Info("Shard proof requested", "shardId", req.ShardID)
+
+	shardPath := new(big.Int).SetInt64(int64(req.ShardID))
+	merkleTreePath, err := pas.parentRoundManager.GetSMT().GetPath(shardPath)
```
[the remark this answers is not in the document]
jait91: GetPath is not called after validateShardID returns error

PR #86, opened 2025-11-17
Dynamic trust base

lploom:


internal/storage/mongodb/trust_base.go
```diff
@@ -0,0 +1,152 @@
...
+		previousTrustBaseV1, ok = previousTrustBase.(*types.RootTrustBaseV1)
+		if !ok {
+			return fmt.Errorf("failed to cast previous trust base to version 1 for epoch %d", epoch)
+		}
+	}
+	trustBaseV1, ok := trustBase.(*types.RootTrustBaseV1)
+	if !ok {
+		return fmt.Errorf("failed to cast provided trust base to version 1 for epoch %d", epoch)
+	}
+	if err := trustBaseV1.Verify(previousTrustBaseV1); err != nil {
```
MastaP: looked into `Verify()`, it checks `PreviousEntryHash`, but does not verify that new trustbase was signed by the previous majority set of validators. This seems like a security issue, when anyone could extend the tb and submit it a new one

PR #88, opened 2025-11-19
Sharding improvements

jait91:
* refactor perf test
* remove unnecessary round timer from round managers
* add `finalized` flag to blocks and add recovery mechanism on startup (retry storing commitments to DB and finalize the block)
* fatal exit on processing errors - exit immediately on round processing failures, recovery handles on restart


cmd/performance-test/types.go
```diff
@@ -0,0 +1,531 @@
...
+
+type MerkleTreePath struct {
+	Root  string               `json:"root"`
+	Steps []api.MerkleTreeStep `json:"steps"`
+}
+
+type shardTarget struct {
+	name      string
+	url       string
+	shardMask int
```
lploom: Should probably be called shardID

PR #98, opened 2025-12-01
Add crash recovery for block finalization

jait91:
Adds finalized flag to blocks and recovery mechanism at startup

### DIFF
--- a/internal/round/batch_processor.go
+++ b/internal/round/batch_processor.go
@@ -551,23 +551,19 @@ func (rm *RoundManager) storeDataParallel(
 	var wg sync.WaitGroup
 
 	if len(smtNodes) > 0 {
-		wg.Add(1)
-		go func() {
-			defer wg.Done()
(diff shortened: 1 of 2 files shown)


internal/round/batch_processor.go
```diff
@@ -509,51 +489,103 @@ func (rm *RoundManager) FinalizeBlock(ctx context.Context, block *models.Block)
...
+	var smtTime, recordsTime time.Duration
 
-	if err := rm.storage.AggregatorRecordStorage().StoreBatch(ctx, records); err != nil {
-		return fmt.Errorf("failed to store aggregator records batch: %w", err)
+	// Run SMT and AggregatorRecords storage in parallel
+	var wg sync.WaitGroup
+
+	if len(smtNodes) > 0 {
+		wg.Add(1)
+		go func() {
```
lploom: Could use Wg.Go()

PR #106, opened 2025-12-30
Dynamic trust base v2

lploom:
set default root epoch to 1


MastaP:
@b3y0urs3lf need to check that aggregator starts with existing trustbase (https://github.com/unicitynetwork/unicity-ids/blob/main/bft-trustbase.testnet.json)

PR #141, opened 2026-04-12
Implement v2 tree and inclusion proof format

### LINKED ISSUE #140: [TGE] Update aggregation tree and inclusion proof formats

jait91:


internal/service/v2_surface_test.go
```diff
@@ -0,0 +1,42 @@
...
+
+	"github.com/unicitynetwork/aggregator-go/pkg/jsonrpc"
+)
+
+func TestRemovedV1MethodsReturnMethodNotFound(t *testing.T) {
+	serverAddr, cleanup := setupMongoDBAndAggregator(t, context.Background())
+	defer cleanup()
+
+	for _, method := range []string{
+		"submit_commitment",
```
MastaP, line 20: This test demonstrates that the three v1 methods are correctly unreachable over HTTP — but the v1 service layer implementing them is still fully present. `internal/service/service.go:26-34` keeps `SubmitCommitment`, `GetInclusionProofV1`, `GetBlockCommitments` on the `Service` interface, and both `AggregatorService` and `ParentAggregatorService` still implement them. Their request/response types remain in `pkg/api/types_v1.go` and `internal/gateway/handlers_v1.go` was only deleted on the gateway side.

The result is a chunk of dead code that can only be exercised by direct Go calls, creating confusion about whether v1 is still supported. Either:

- **Finish the removal**: drop the three methods from the `Service` interface, delete the `AggregatorService` implementations (lines 177-268 and 325-403 in `service.go`), delete the parent-mode error-return stubs at `parent_service.go:242-254` and `284-286`, and remove the associated `api.SubmitCommitmentRequest` / `GetInclusionProofRequestV1` / `GetBlockCommitmentsRequest` types in `pkg/api/types_v1.go`.
- **Or re-register the handlers** in `server.go:164` for a deprecation period, and document the sunset date.

Leaving the current half-state makes it easy for a future change to accidentally re-introduce a v1 RPC path by re-registering a handler that still has a backing implementation.

internal/smt/smt.go
```diff
@@ -308,8 +330,13 @@ func (l *LeafBranch) isLeaf() bool {
 
 // NewNodeBranch creates a regular node branch
 func newNodeBranch(path *big.Int, left, right branch) *NodeBranch {
+	return newNodeBranchWithDepth(path, left, right, path.BitLen()-1)
+}
+
+func newNodeBranchWithDepth(path *big.Int, left, right branch, depth int) *NodeBranch {
 	return &NodeBranch{
 		Path:   new(big.Int).Set(path),
+		Depth:  uint8(depth),
```
MastaP, line 353: Latent `uint8` depth overflow with no bounds check. `NodeBranch.Depth` is `uint8` and `calculateHash` hashes `AddData([]byte{n.Depth})` as a single byte (`smt.go:389`). For the current `StateTreeKeyLengthBits = 256` the max valid depth is 255, so it just happens to fit. But if any future caller bumps the key length — or if the parent tree is ever constructed with `shardIDLength` bits that push total depth past 256 — `uint8(depth)` will silently wrap, and two structurally different nodes at depths `d` and `d+256` will produce the same hash. Both `newNodeBranchWithDepth` (here) and `newRootBranch` at line 351 have the same issue.

`generateInclusionCert` at line 565 already has the defensive check:

```go
if depth < 0 || depth >= api.StateTreeKeyLengthBits {
    return fmt.Errorf("smt: invalid branch depth %d", depth)
}
```

The same guard belongs in the constructor so we fail fast at construction rather than producing a silently-wrapped tree that passes through until proof generation. Suggest:

```go
func newNodeBranchWithDepth(path *big.Int, left, right branch, depth int) *NodeBranch {
    if depth < 0 || depth > 255 {
        panic(fmt.Sprintf("smt: node depth %d out of range [0,255]", depth))
    }
    return &NodeBranch{...}
}
```

(Panic is consistent with the existing `panic("SMT key length must be positive")` at `smt.go:71`.)

internal/smt/smt.go
```diff
@@ -290,9 +309,12 @@ func (l *LeafBranch) calculateHash(hasher *api.DataHasher) []byte {
...
-	hasher.Reset().AddData(api.CborArray(2)).
-		AddCborBytes(pathBytes).AddCborBytes(l.Value)
+	// v2 leaf hashing with domain separation:
+	// H(0x00 || key || value)
+	hasher.Reset().
+		AddData([]byte{0x00}).
+		AddData(l.Key).
+		AddData(l.Value)
 	hasher.SumRaw(l.rawHash[:0])
 	l.hashSet = true
```
MastaP, line 330: Read-side data race on the inline hash cache. The new `rawHash [smtCachedHashBytes]byte` + `hashSet bool` fields on both `LeafBranch` and `NodeBranch` are mutated by `calculateHash` without synchronization, but `ThreadSafeSMT` public reads (`GetRootHash`, `GetPath`, `GetInclusionCert`, `GetRootHashRaw`) only acquire `rwMux.RLock()`. Two concurrent readers that hit an unprimed subtree will race on the writes to `rawHash[:]` and `hashSet = true`:

- Reader A sees `hashSet == false`, begins computing, starts writing `rawHash`.
- Reader B sees `hashSet == true` (already set by A under x86 TSO) but observes a partially-written `rawHash` on non-TSO platforms. Result: wrong root hash returned to a client.

In the current flow this is usually avoided in practice because `processRound` (`round_manager.go:491`) calls `Snapshot.GetRootHashRaw()` before `Commit`, which cascades through every reachable node and primes the cache; after `Commit` the swapped root has all cached hashes set before any concurrent RLock reader can reach it. But:

1. The invariant is implicit — nothing enforces "the cache must be primed before the tree is visible under RLock."
2. `GetInclusionCert` on the main SMT (not just snapshots) is called from `GetInclusionProofV2`, and the priming comment at line 534-537 only primes inside that call. A concurrent `GetPath` from v1 on the same tree at the same time could still race if the priming happens to be in flight.
3. `go test -race` would flag this on a crafted test that invokes two readers on an unprimed tree.

Options:

- Use `sync/atomic` (`atomic.Bool` for `hashSet`, or a single `atomic.Pointer[[32]byte]` for the whole cache slot) to give readers a happens-before guarantee on the write.
- Acquire the write lock in `calculateHash` (heavy, and defeats the point of the inline cache).
- Document the invariant explicitly in `calculateHash` and add a fail-fast check: callers must have primed the cache before taking RLock.

The atomic route is the cheapest and preserves the performance win from the inline array. Happy to sketch a patch if useful.

internal/smt/smt.go
jait91: Added explicit depth bounds checks in both node/root branch constructors so invalid depths fail fast instead of silently truncating into uint8.

internal/smt/smt.go
jait91: Hash caches are now primed at construction time as well as after writes/commit, so read paths under RLock no longer populate cache state on a fresh wrapped tree.

internal/gateway/server.go
```diff
@@ -164,20 +161,15 @@ func (s *Server) setupJSONRPCHandlers() {
 		s.rpcServer.RegisterMethod("get_shard_proof", s.handleGetShardProof)
 	} else {
 		// Standalone mode handlers (default)
-		s.rpcServer.RegisterMethod("submit_commitment", s.handleSubmitCommitment)         // v1
-		s.rpcServer.RegisterMethod("certification_request", s.handleCertificationRequest) // v2
-
-		s.rpcServer.RegisterMethod("get_inclusion_proof", s.handleGetInclusionProofV1)    // v1
+		s.rpcServer.RegisterMethod("certification_request", s.handleCertificationRequest)
 		s.rpcServer.RegisterMethod("get_inclusion_proof.v2", s.handleGetInclusionProofV2) // v2
 	}
```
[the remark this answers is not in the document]
jait91: Handled in the sharding PR

internal/round/batch_processor.go
```diff
@@ -596,12 +587,19 @@ func (rm *RoundManager) convertLeavesToNodes(leaves []*smt.Leaf) []*models.SmtNo
...
+			rm.logger.Error("failed to convert leaf path to SMT storage key",
+				"path", leaf.Path.String(),
+				"error", err.Error())
+			continue
+		}
 		key := api.NewHexBytes(keyBytes)
 		value := api.NewHexBytes(leaf.Value)
-		smtNodes[i] = models.NewSmtNode(key, value)
+		smtNodes = append(smtNodes, models.NewSmtNode(key, value))
 	}
```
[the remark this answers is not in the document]
jait91: Leaf persistence conversion now returns an error instead of skipping bad entries, and block finalization aborts on conversion failure

internal/service/v2_surface_test.go
jait91: Agreed it should be cleaned up but will do in a follow-up PR after v2 is fully rolled out

MastaP (commented):
Thanks for the v2 rework — existing reviewer threads (SMT hash-cache race, uint8 depth overflow, silent `convertLeavesToNodes` skip, child-mode `get_inclusion_proof.v2`, v1 Service-method cleanup, `TestShardingE2E` skip) already look addressed; skipping those to avoid noise. A few additional observations on the current tip:

### Potential issues

- **Child-mode proof composition doesn't assert root coherence.** `GetInclusionProofV2` (`internal/service/service.go`) calls `GetLatestByRootHash(rootHashRaw)` and later `smtInstance.GetRootHashRaw()` for `ComposeInclusionCert` under `FinalizationReadLock`. Today the lock makes those observations consistent, but there's no explicit assertion that the two raw roots match — if the locking invariant ever loosens, composition could silently mix a child cert from root A with a parent fragment bound to root B. `ThreadSafeSMT.GetShardInclusionFragmentWithRoot` already atomically snapshots fragment+root under one RLock; mirroring that with a `GetInclusionCertWithRoot` would make the invariant local to the API and survive future refactors.

- **`byte(d)` cast in `InclusionCert.Verify`.** `pkg/api/inclusion_cert.go:~117` hashes `{0x01, byte(d)}` where `d` loops from `maxDepth - 1 = 255` down. Today `maxDepth == StateTreeKeyLengthBits == 256`, so the cast is safe — but if the key length is ever widened, `byte(d)` silently wraps and two structurally different nodes collide. Either reject depths > 255 explicitly or document the wire cap on `StateTreeKeyLengthBits` at the type level.

- **`ParentBlockNumber == 0` sentinel is ambiguous.** `validateBlockForMode` in `internal/round/batch_processor.go` treats zero as "missing parent block number." Parent block indices start at 1 in practice, but `uint64(0)` is also the zero value after a decode bug or a test that forgot to set it. Consider `*uint64` or a paired `HasParentMetadata` flag to distinguish "unset" from "legitimately zero."

- **`InclusionCert.MarshalBinary` can't fail.** The `(error)` return is always `nil`. Same for `ExclusionCert.MarshalBinary`. Either keep for `encoding.BinaryMarshaler` compatibility (fine — just add a one-line comment) or drop the error.

- **`RootShardInclusionProof.IsValid` hides UC decode errors.** In `pkg/api/types.go`, `ucInputRecordHashRaw` failures collapse into `false`, which `pollForParentProof` then logs as a plain retry. A malformed UC from a misbehaving parent will loop silently. Consider `(bool, error)` so the poll loop can log at Warn when decoding fails (distinct from "not ready yet").

### Style / maintainability

- **Depth arithmetic duplicated.** `depthOffset + (commonPath.BitLen() - 1)` appears 4× in `internal/smt/smt.go` `AddLeaf`/`buildTree`. A small helper (`nodeDepthAt(offset, commonPath)`) would make the depth math easier to audit now that it's load-bearing for the v2 hash.

- **`selectShardIndex` fallback comment.** `cmd/performance-test/main.go:~196` says "Fallback only: keep distribution aligned with LSB-first key layout" but only looks at `keyBytes[0]`. Worth stating that this branch only fires when the request didn't already match any configured shard, so readers don't mistake it for routing logic.

- **`pathToBitmapAndSiblings` test helper is scaffolding.** `internal/smt/golden_vectors_test.go` contains a helper that reverses between leaf-to-root and root-to-leaf sibling order for cross-format parity, with a comment mentioning "Rugregator" (typo → "Aggregator"). Once the legacy `MerkleTreePath` type is removed, this helper can go too.

- **`AddPreHashedLeaf` was previously a silent no-op.** Good catch that this now actually delegates to `AddLeaf`; the new `TestThreadSafeSMT_AddPreHashedLeaf_StoresChildRoot` pins the behavior. Worth confirming nothing else in tree construction relied on the old broken semantics (a grep suggests no).

### Other notes
- Test coverage is strong: `golden_vectors_test.go`, `inclusion_cert_test.go` (523 lines of wire + verify edge cases), `yellowpaper_hash_semantics_test.go`, `thread_safe_smt_test.go`, the new child-mode compose tests in `service_test.go`, and `state_id_bitorder_test.go` all look like solid regression protection against bit-order / wire-shape drift.
- Security-wise, removing `JoinPaths` (string-compared "child root must match first parent step") in favor of typed binary `InclusionCert.Verify` bound strictly to `UC.IR.h` is a clear net improvement.
- `ensureHashes()` on every write is O(dirty path) not O(N) thanks to the `hashSet` short-circuit, so the race fix shouldn't move `make performance-test` numbers meaningfully — worth a run on the sharded topology to confirm before merge.

README.md
```diff
@@ -203,57 +203,24 @@ Submit a state transition request to the aggregation layer with cryptographic va
...
-      58 41                                  #     bytes(65)
-         03ce4ef0fe3b4f53f5264daee6930c5e    #       "\x03\xceN\xf0\xfe;OS\xf5&M\xae\xe6\x93\x0c^"
-         7a3b60f4dfd102b4d8f2420d8bbba17e    #       "z;`\xf4\xdf\xd1\x02\xb4\xd8\xf2B\r\x8b\xbb\xa1~"
-         446b0f855ad402437f14d00c1f27752e    #       "Dk\x0f\x85Z\xd4\x02C\x7f\x14\xd0\x0c\x1f\'u."
-         9aa802301ca42a57a80cb1f6f57e03eb    #       "\x9a\xa8\x020\x1c\xa4*W\xa8\x0c\xb1\xf6\xf5~\x03\xeb"
-         00                                  #       "\x00"
-   f5                                        #   true, simple(21)
-   00                                        #   unsigned(0)
-```
+Params is a hex-encoded `api.CertificationRequest` CBOR payload. The payload contains raw 32-byte values for `stateId`, `sourceStateHash`, and `transactionHash`; these fields do not include algorithm prefixes.
```
MastaP, line 211: The previous README had a concrete CBOR example with a fully decoded breakdown — this PR replaces it with a generic "<hex-encoded CertificationRequest CBOR>" placeholder. Downstream SDK authors lose a useful reference for what a v2 payload actually looks like on the wire. Consider keeping a minimal v2 hex string + decoded form so consumers can sanity-check their encoders (especially since the algorithm-prefix bytes are now gone — that's exactly the kind of thing readers want to see in concrete form).

pkg/api/inclusion_cert.go
```diff
@@ -0,0 +1,250 @@
...
+	if hasher == nil {
+		return fmt.Errorf("%w: %d", ErrCertUnknownAlgo, algo)
+	}
+
+	// Walk depths from deepest to shallowest, consuming siblings from
+	// the end of the slice. Depths with bitmap bit clear are skipped
+	// (unary passthrough or off-path).
+	h := append([]byte(nil), startHash...)
+	j := len(siblings)
+	for d := maxDepth - 1; d >= 0; d-- {
```
MastaP, line 139: The sibling-consumption order is correct but not obvious on first read. The wire is documented as "root-to-leaf" (shallowest first), but the verifier walks `d = 255 → 0` while consuming `siblings[len-1] → siblings[0]`. The two reversals cancel out, which is what you want — but the only way to verify that is to mentally trace through. A 4-line ASCII diagram in this function's doc, e.g.

```
//   wire:    s_1 (shallowest) ... s_n (deepest)
//   walk:    d=255 (deepest) ... d=0 (shallowest)
//   consume: siblings[n-1]   ... siblings[0]
//   → at deepest depth we take the last sibling = deepest sibling ✓
```

would save the next reader several minutes of pencil work.

pkg/api/types.go
```diff
@@ -218,15 +301,25 @@ func (h *HealthStatus) AddDetail(key, value string) {
...
 }
 
-func (r *RootShardInclusionProof) IsValid(shardRootHash string) bool {
-	return r.MerkleTreePath != nil && len(r.UnicityCertificate) > 0 &&
-		len(r.MerkleTreePath.Steps) > 0 && r.MerkleTreePath.Steps[0].Data != nil && *r.MerkleTreePath.Steps[0].Data == shardRootHash
+func (r *RootShardInclusionProof) IsValid(shardID ShardID, keyLength int, shardRootHash HexBytes) bool {
+	if r == nil || len(r.UnicityCertificate) == 0 || r.ParentFragment == nil {
+		return false
+	}
+	rootRaw, err := ucInputRecordHashRaw(r.UnicityCertificate)
```
MastaP, line 308: `ucInputRecordHashRaw` failures here collapse into `return false`, which the caller (`pollForParentProof`) treats as a normal "not ready yet" retry. A malformed UC from a misbehaving parent will then loop silently with no signal in the logs.

Consider returning `(bool, error)` so callers can distinguish "valid UC but proof not yet anchored" from "UC failed to decode" and log the latter at Warn — otherwise a real parent-side encoding bug looks identical to a slow round.

internal/smt/smt.go
```diff
@@ -550,47 +763,52 @@ func (smt *SparseMerkleTree) buildTree(branch branch, remainingPath *big.Int, va
...
 		oldBranchPath := new(big.Int).Rsh(leafBranch.Path, uint(commonPath.BitLen()-1))
-		oldBranch := newLeafBranch(oldBranchPath, leafBranch.Value)
+		oldBranch := newLeafBranchWithKey(oldBranchPath, leafBranch.Key, leafBranch.Value)
 
 		// TypeScript: remainingPath >> commonPath.length
 		newBranchPath := new(big.Int).Rsh(remainingPath, uint(commonPath.BitLen()-1))
-		newBranch := newLeafBranch(newBranchPath, value)
+		newBranch := newLeafBranchWithKey(newBranchPath, leafKey, value)
+
+		nodeDepth := depthOffset + (commonPath.BitLen() - 1)
```
MastaP, line 772: `depthOffset + (commonPath.BitLen() - 1)` appears 4× in `buildTree` — here, line 790, line 799, and the recursion at line 801/807. Now that depth is load-bearing for v2 hashes (`H(0x01 || depth || left || right)`) and the `uint8` overflow guard panics in the constructor, an arithmetic slip in any of these copies would crash at construction time rather than producing a silently-wrapped hash, which is good — but it still means the math should live in one place.

A tiny helper like

```go
func nodeDepthAt(depthOffset int, commonPath *big.Int) int {
    return depthOffset + commonPath.BitLen() - 1
}
```

would make the depth wiring trivially auditable and remove a place where a future refactor could drift. Not a correctness issue today; purely maintainability.

MastaP (commented):
Four follow-up nits beyond the items already resolved in the threads above. None are blocking — feel free to handle in a follow-up.

PR #142, opened 2026-04-14
Implement sharded v2 inclusion proof composition

jait91:


MastaP (commented):
Seven inline comments below covering items 5, 7-12 from a thorough review. Items 1-4 (self-verify CPU cost on read path, v1-in-child-mode hard break, child-mode BlockNumber semantics change, `ParentBlockNumber==0` overload) are more consequential and I'll raise those separately.

**Item 6 (dead code, doesn't map to a diff hunk so no inline)**: `smt.JoinPaths` at `internal/smt/smt.go:953` has no production caller left after this PR — the old child-mode path join in `service.go` was deleted. The only remaining reference is `TestJoinPaths` in `smt_test.go:1384`. Delete both the function and the test to reduce maintenance surface.

internal/models/block.go
```diff
@@ -112,33 +125,33 @@ func (bb *BlockBSON) FromBSON() (*Block, error) {
...
+		UnicityCertificate:  unicityCertificate,
+		ParentFragment:      parentFragment,
+		ParentBlockNumber:   parentBlockNumber,
+		Finalized:           bb.Finalized,
 	}, nil
 }
 
 // NewBlock creates a new block
-func NewBlock(index *api.BigInt, chainID string, shardID api.ShardID, version, forkID string, rootHash, previousBlockHash, uc api.HexBytes, parentMerkleTreePath *api.MerkleTreePath) *Block {
+func NewBlock(index *api.BigInt, chainID string, shardID api.ShardID, version, forkID string, rootHash, previousBlockHash, uc api.HexBytes) *Block {
```
MastaP, line 135: **NewBlock() no longer takes parent fields; they're assigned after construction.**

In `internal/round/batch_processor.go:215-216` the caller does:
```go
block := models.NewBlock(...)
block.ParentFragment = proof.ParentFragment
block.ParentBlockNumber = proof.BlockNumber
```

A future caller in child mode could forget either assignment and silently produce a broken block — the error only surfaces at *serving* time via `proofBundleBlockNumber`. Consider either:

1. A dedicated `NewChildBlock(...)` constructor that takes `parentFragment` and `parentBlockNumber` as required args, or
2. A finalization-time invariant check so malformed child blocks can never reach storage in the first place.

Fail-fast at write time is better than fail-forever at read time.

internal/smt/smt.go
```diff
@@ -599,27 +648,37 @@ func (smt *SparseMerkleTree) generateInclusionCert(hasher *api.DataHasher, key [
 			child = node.Right
 		}
 
+		childHash := child.calculateHash(hasher)
+		if childHash == nil {
+			return nil, nil
+		}
+
 		sibHash := sibling.calculateHash(hasher)
+		if sibHash == nil {
```
MastaP, line 660: **Silent sibling-skip branch needs an invariant comment and a direct unit test.**

```go
sibHash := sibling.calculateHash(hasher)
if sibHash == nil {
    return smt.generateInclusionCertWithLeafValue(hasher, key, child, cert)
}
```

If this fires *after* some siblings have already been recorded in `cert`, the resulting cert has those siblings but is silently missing the bitmap bit for the current depth. I can't convince myself from the code alone that the resulting cert is well-formed.

Two requests:
1. Add a comment explaining *when* `sibling.calculateHash` returns nil (empty subtree? uninitialized child-mode placeholder?) and why skipping the depth preserves proof correctness.
2. Add a unit test that triggers this branch with non-empty prior siblings and asserts `composed.Verify(...)` against the real root succeeds.

Related: the nil-leaf-key handling at line 626 (`if leaf.Key != nil && !bytes.Equal(...)`) relaxed the old "leaf not found" error. That semantic change should also get a doc comment.

pkg/api/types.go
```diff
@@ -138,9 +139,51 @@ type InclusionProofV2 struct {
...
+		return errors.New("parent fragment shard leaf value does not match expected child root")
+	}
+
+	var cert InclusionCert
+	if err := cert.UnmarshalBinary(f.CertificateBytes); err != nil {
+		return fmt.Errorf("failed to decode parent fragment cert: %w", err)
+	}
+
+	path := big.NewInt(int64(shardID))
+	if path.Sign() <= 0 || path.BitLen() <= 1 {
```
MastaP: **Redundant pre-check — `PathToFixedBytes` already validates this more strictly.**

```go
if path.Sign() <= 0 || path.BitLen() <= 1 {
    return fmt.Errorf("invalid shard ID %d", shardID)
}
if keyLength <= 0 {
    return fmt.Errorf("keyLength must be positive, got %d", keyLength)
}
```

`api.PathToFixedBytes` (state_id.go:73-82) already enforces `path.Sign() > 0`, `keyLengthBits > 0`, and `path.BitLen()-1 == keyLengthBits`. The pre-check here is weaker (it doesn't verify the bit length *matches* keyLength) and produces a worse error message.

Suggest deleting lines 169-174 entirely and letting `PathToFixedBytes` produce the error with full context.

internal/service/parent_service.go
```diff
@@ -171,29 +170,21 @@ func (pas *ParentAggregatorService) GetShardProof(ctx context.Context, req *api.
 
 	pas.logger.WithContext(ctx).Debug("Shard proof requested", "shardId", req.ShardID)
 
-	shardPath := new(big.Int).SetInt64(int64(req.ShardID))
-	merkleTreePath, err := pas.parentRoundManager.GetSMT().GetPath(shardPath)
+	parentFragment, rootHashRaw, err := pas.parentRoundManager.GetSMT().GetShardInclusionFragmentWithRoot(req.ShardID)
```
MastaP, line 176: **Atomic fragment+root read, non-atomic block lookup — document the ordering invariant.**

`GetShardInclusionFragmentWithRoot` grabs the fragment and SMT root atomically under RLock, then the lock is released before `BlockStorage().GetLatestByRootHash` runs. Correctness depends on the invariant that *block storage commit is ordered before SMT commit* (the snapshot pattern from CLAUDE.md's SMT notes).

Two asks:
1. Add a comment here stating that invariant explicitly, so a future reader understands why this isn't a TOCTOU bug.
2. Is there a startup/recovery window where the in-memory SMT has a root for which the corresponding block hasn't been reconstructed yet? If so, `GetLatestByRootHash` could transiently return nil on the leader shortly after takeover. Worth verifying there's a test that covers this (or at least that `parentRoundManager.IsReady()` gates against it).

internal/signing/certification_request_validator.go
```diff
@@ -195,33 +194,18 @@ func (v *CertificationRequestValidator) ValidateShardID(stateID api.StateID) err
 	return nil
 }
 
-// verifyShardID Checks if commitmentID's least significant bits match the shard bitmask.
+// verifyShardID checks whether the raw SMT key bits of commitmentID match the
+// configured shard prefix under the SMT's LSB-first byte layout.
 func verifyShardID(commitmentID string, shardBitmask int) (bool, error) {
```
MastaP: **`verifyShardID` is still duplicated between v1 and v2 validators.**

The core bit-matching loop was correctly factored into `api.MatchesShardPrefix`, but the two wrappers remain:
- `internal/signing/certification_request_validator.go:199` — requires exactly 32 bytes.
- `internal/signing/v1/commitment_validator.go:226` — accepts 34 bytes and strips the 2-byte imprint prefix.

They differ only in how they normalize the input. A single helper like `api.MatchesShardPrefixFromImprintOrRaw(hex string, shardMask int) (bool, error)` that handles both input shapes would eliminate both wrappers. Not a blocker for this PR — flagging as cleanup.

internal/models/block.go
```diff
@@ -48,13 +50,17 @@ func (b *Block) ToBSON() (*BlockBSON, error) {
...
-		merkleTreePathJson, err := json.Marshal(b.ParentMerkleTreePath)
+	var parentFragment string
+	if b.ParentFragment != nil {
+		parentFragmentJSON, err := json.Marshal(b.ParentFragment)
 		if err != nil {
-			return nil, fmt.Errorf("failed to marshal parent merkle tree path: %w", err)
+			return nil, fmt.Errorf("failed to marshal parent fragment: %w", err)
 		}
-		merkleTreePath = api.NewHexBytes(merkleTreePathJson).String()
+		parentFragment = api.NewHexBytes(parentFragmentJSON).String()
```
MastaP: **Double-encoding `ParentFragment` in BSON.**

```go
parentFragmentJSON, err := json.Marshal(b.ParentFragment)
...
parentFragment = api.NewHexBytes(parentFragmentJSON).String()
```

The fragment's own fields (`HexBytes`) are already hex strings in JSON, so we're storing *hex-encoded JSON containing hex strings* — roughly 2× the necessary bytes, plus two decode passes on read. Storing `ParentFragment` as a BSON subdocument (`ParentFragment *ParentInclusionFragmentBSON`) would cut storage ~50% and avoid the JSON→hex round-trip.

This pattern was inherited from the deleted `ParentMerkleTreePath` code and is pure cleanup — tests already pass either way.

cmd/performance-test/types.go
```diff
@@ -111,6 +112,7 @@ type ShardClient struct {
 type proofJob struct {
 	shardIdx  int
 	requestID string
```
MastaP: **`proofJob.requestID` is now redundant.**

```go
type proofJob struct {
    shardIdx  int
    requestID string
    request   *api.CertificationRequest
}
```

After adding `request`, `requestID` is just `request.StateID.String()`. Drop the separate field and update the call sites that reference it — small cleanup, reduces one source of drift.

internal/models/block.go
jait91: Fixed. Child-mode blocks now use a dedicated NewChildBlock(...) constructor

internal/smt/smt.go
jait91: Added invariant comments for both cases + test

pkg/api/types.go
jait91: fixed

internal/service/parent_service.go
jait91: Added the ordering-invariant comment at the lookup site and added the `ParentRoundManager.IsReady()` gate so shard proofs are not served before the parent manager is initialized.

internal/signing/certification_request_validator.go
jait91: I kept the shared matcher in api, but left the 34-byte normalization in the V1 validator on purpose. The remaining wrapper cleanup can be done later with the broader V1 cleanup.

internal/models/block.go
jait91: Fixed. `ParentFragment` is now stored as a BSON subdocument instead of hex-encoded JSON.

cmd/performance-test/types.go
jait91: removed unneeded field

MastaP (approved):

PR #143, opened 2026-04-15
Cbor tags

### LINKED ISSUE #138: [TGE] Update aggregator implementation to use the new type tags
Protocol elements need to be wrapped in type tags and version number fields added like it's already done in the BFT node implementation.

MastaP:


internal/gateway/docs_test.go
```diff
@@ -18,7 +18,7 @@ import (
 func TestDocumentationExamplePayload(t *testing.T) {
 	// Extract the example payload from the documentation
 	// This is the exact payload shown in the docs
-	exampleJSON, err := hex.DecodeString("83582032bbbc306e3a101e8fa1e4349631864957c7d84167238c85cba972943ce05ccd8483014101582102cbbbe7dc6d51dea5c5fb4d7e7da3416e5914b989c303399f31b51db090981cfa58208126936ae7bcd660a93368a8c83951e01ccbcd4af093769b1cc14f942e2a9ca85820ad8f039daf3827446a0af7ccf31b438aea079440406a07cca7374b52b4e84c2c584177ff8a78a8e59f71d03e687fa14851327babb6a4f5cafcdfc56b16b2267284247734523df030465e249734aab146ae18d65bb3c7aee570c4f53e56e779e0c4200100")
+	exampleJSON, err := hex.DecodeString("d9987684015820f6a9010354c4359cbe9d63892684bbff6bd54ef86e9dd98a155a1a32716a0247d998778501d9987883014101582102cbbbe7dc6d51dea5c5fb4d7e7da3416e5914b989c303399f31b51db090981cfa58208126936ae7bcd660a93368a8c83951e01ccbcd4af093769b1cc14f942e2a9ca85820ad8f039daf3827446a0af7ccf31b438aea079440406a07cca7374b52b4e84c2c584177ff8a78a8e59f71d03e687fa14851327babb6a4f5cafcdfc56b16b2267284247734523df030465e249734aab146ae18d65bb3c7aee570c4f53e56e779e0c4200100")
```
ahtotruu: Can we fix the misleading variable name here? :)

pkg/api/certification_request.go
```diff
@@ -23,6 +24,22 @@ type CertificationRequest struct {
 	AggregateRequestCount uint64
 }
 
+func (c *CertificationRequest) GetVersion() types.Version { return 1 }
+
+func (c *CertificationRequest) MarshalCBOR() ([]byte, error) {
+	type alias CertificationRequest
```
ahtotruu, line 35: Not sure of the benefit of the type alias here, because there's still "hardcoded" `CertificationRequestTag` next to the "generalized" `alias` below.

pkg/api/certification_request.go
```diff
@@ -23,6 +24,22 @@ type CertificationRequest struct {
...
+	cp := *c
+	if cp.Version == 0 {
+		cp.Version = c.GetVersion()
+	}
+	return types.Cbor.MarshalTaggedValue(CertificationRequestTag, (*alias)(&cp))
+}
+
+func (c *CertificationRequest) UnmarshalCBOR(data []byte) error {
+	type alias CertificationRequest
```
ahtotruu, line 44: Same here.

pkg/api/types.go
```diff
@@ -133,12 +133,29 @@ type GetInclusionProofResponseV2 struct {
 //
 // See docs/inclusion-proof-wire.md for the frozen specification.
 type InclusionProofV2 struct {
-	_                  struct{}           `cbor:",toarray"`
+	_                  struct{} `cbor:",toarray"`
+	Version            types.Version
 	CertificationData  *CertificationData `json:"certificationData"`
```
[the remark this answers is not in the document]
ahtotruu: This looks like a valid point to me.

pkg/api/certification_request.go
MastaP: it's to avoid infinite recursion on MarshalCBOR, `alias` type does not have it

jait91 (approved):

ahtotruu (approved):

PR #146, opened 2026-04-22
Bft sharding

### LINKED ISSUE #144: Add BFT-based sharding support
Implement the ability to run the aggregator as one shard in a sharded partition under the BFT core. This includes:
- configuring the shard ID when the aggregator starts;
- only accepting the aggregation requests for state identifiers matching the shard ID;
- including the shard ID in the certification requests sent to the BFT cluster;
- including the shard tree certificate in the responses sent to clients.

There is currently no requirement for combining this sharding mode with the one implemented in #54 (to deploy a sharded aggregator, the operator must choose one of the two sharding modes).

jait91:


MastaP (commented):
Code review — items not already raised by gemini-code-assist.

internal/service/service.go
```diff
@@ -167,7 +171,7 @@ func NewAggregatorService(cfg *config.Config,
 		roundManager:                  roundManager,
 		leaderSelector:                leaderSelector,
 		commitmentValidator:           signingV1.NewCommitmentValidator(cfg.Sharding),
-		certificationRequestValidator: signing.NewCertificationRequestValidator(cfg.Sharding),
+		certificationRequestValidator: signing.NewCertificationRequestValidator(cfg.Sharding, bftShardID),
```
MastaP, line 174: This wires `bftShardID` into the v2 validator, but the v1 `commitmentValidator` (built one line above) still uses `signingV1.NewCommitmentValidator`, whose `ValidateShardID` at `internal/signing/v1/commitment_validator.go:210-213` short-circuits `return nil` for anything that is not child mode. Net effect: `submit_commitment` in bft-shard mode accepts stateIDs that belong to *any* shard. The record lands in storage and the SMT, but is then unreachable: `GetInclusionProofV1` errors out in bft-shard mode (service.go:333-336) and `GetInclusionProofV2` rejects it because `record.Version != 2` (line 468). More importantly, because the v1 submit path does not enforce prefix ownership, nothing prevents the same stateID from being accepted on two different bft-shard aggregators in the same partition — breaking the single-admission invariant bft-shard is supposed to provide. Either teach `signingV1.CommitmentValidator.ValidateShardID` about bft-shard (mirror the `bftShardID.Comparator()` check added to the v2 validator), or reject `submit_commitment` outright at the service layer when `Mode == ShardingModeBFTShard`.

internal/service/service.go
```diff
@@ -159,6 +159,10 @@ func NewAggregatorService(cfg *config.Config,
 	leaderSelector LeaderSelector,
 	receiptSigner *signing.ReceiptSigner,
 ) *AggregatorService {
+	var bftShardID types.ShardID
+	if cfg.BFT.ShardConf != nil {
+		bftShardID = cfg.BFT.ShardConf.ShardID
```
MastaP, line 164: The validator captures `cfg.BFT.ShardConf.ShardID` by value once at construction. Trust bases already have a dynamic update path (`PutTrustBase`); if the BFT shard conf ever grows a similar live update (e.g. epoch-driven shard reconfig), this copy stays stale while still gating every `certification_request`. A TODO here would keep whoever adds dynamic shard-conf reconfig from missing it; alternatively, store `cfg` on the validator and read `ShardConf.ShardID` lazily.

internal/config/config.go
```diff
@@ -444,6 +452,26 @@ func (c *Config) Validate() error {
...
 
+	if c.Sharding.Mode == ShardingModeBFTShard {
+		if !c.BFT.Enabled {
+			return errors.New("bft-shard mode requires BFT_ENABLED=true")
+		}
+		if c.BFT.ShardConf == nil {
+			return errors.New("bft-shard mode requires BFT_SHARD_CONF_FILE")
+		}
+		if c.BFT.ShardConf.ShardID.Length() == 0 {
+			return errors.New("bft-shard mode requires a non-empty shard ID; use standalone for single-shard partitions")
```
MastaP, line 463: Consider also rejecting `c.Sharding.Child.ShardID != 0` in the bft-shard block. Standalone mode already guards against this in `ShardingConfig.Validate()`, but bft-shard silently ignores the legacy app-shard integer. If an operator switches `SHARDING_MODE=standalone → bft-shard` without clearing `SHARDING_CHILD_SHARD_ID`, the residual value stays on the config struct; today it is only hidden from `buildShardingHealth` thanks to the mode guard, and any future reader that forgets that guard will pick it up. Symmetric guard makes the misconfig loud at startup rather than silent.

cmd/performance-test/main.go
```diff
@@ -88,21 +119,49 @@ func getEnvShardTargets() []shardTarget {
...
+				url:       url,
+				shardMask: mask,
+			})
+		case shardingModeBFT:
+			// Validate the bit string is only 0s and 1s. Empty string means
+			// "single-shard partition" (accepts every stateID).
+			for _, r := range field {
+				if r != '0' && r != '1' {
+					log.Printf("Warning: invalid shard bits '%s' (bft mode expects binary 0/1 string)", field)
+					field = "__invalid__"
```
MastaP, line 149: Using a magic string as a control-flow sentinel is fragile. A simple boolean makes the intent obvious and does not overload `field`:

```go
valid := true
for _, r := range field {
    if r != '0' && r != '1' {
        log.Printf(...)
        valid = false
        break
    }
}
if !valid {
    continue
}
```

internal/bft/client.go
```diff
@@ -301,9 +301,26 @@ func (c *BFTClientImpl) handleCertificationResponse(ctx context.Context, cr *cer
 		return fmt.Errorf("got CertificationResponse for a wrong shard %s - %s", cr.Partition, cr.Shard)
 	}
 
+	if err := validateUCShardBinding(&cr.UC, c.ShardID()); err != nil {
```
MastaP: The comment on `pkg/api/types.go:396-397` notes that `uc.Verify` currently ignores its shardID parameter, which justifies this extra check today. When bft-go-base makes `uc.Verify` enforce shard equality, this call becomes redundant but will not show up as a diff anywhere. A short `TODO(bft-core): drop once uc.Verify enforces shardID` on this line (and on the matching check in `pkg/api/types.go:398-401`) would keep the two from silently outliving their purpose.

internal/smt/smt.go
```diff
@@ -560,9 +560,13 @@ func (smt *SparseMerkleTree) GetInclusionCert(key []byte) (*api.InclusionCert, e
 	_ = smt.root.calculateHash(hasher)
 
 	var cert api.InclusionCert
-	if _, err := smt.generateInclusionCertWithLeafValue(hasher, key, smt.root, &cert); err != nil {
+	leafValue, err := smt.generateInclusionCertWithLeafValue(hasher, key, smt.root, &cert)
+	if err != nil {
 		return nil, err
 	}
+	if len(leafValue) == 0 {
+		return nil, fmt.Errorf("smt: leaf not found for key %x", key)
```
MastaP, line 570: Good fix — the prior silent empty-cert behavior was easy to misread as success. Worth surfacing the contract in this method's godoc (lines 546-551): today it says 'Returns an error if no leaf exists at the key' which reads as leaf-at-key mismatch, but this path now also fires for placeholder leaves with nil value in parent-mode trees (as pinned by `TestGetInclusionCert_PlaceholderLeaf`). A one-line note that 'a placeholder leaf with no value counts as not-found' would save a future reader from having to walk `generateInclusionCertWithLeafValue` to understand when this fires.

pkg/api/types.go
```diff
@@ -318,18 +323,36 @@ func (c *GetInclusionProofResponseV2) UnmarshalJSON(data []byte) error {
...
+	ShardConfHash   []byte
+}
+
+// Verify checks a v2 inclusion proof end-to-end against the outer
+// CertificationRequest and VerifierContext: local SMT path, UnicityCertificate
+// (shard tree → unicity tree → seal), and ShardTreeCertificate.Shard equality.
+//
+// The nil-guard error strings below are part of the public contract so
+// reference verifiers in other languages can pin them.
+func (p *InclusionProofV2) Verify(v2 *CertificationRequest, vctx *VerifierContext) error {
```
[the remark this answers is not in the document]
jait91: Fixed the duplication by extracting the SMT-only check into a shared internal helper used by both CLIs

cmd/commitment/main.go
```diff
@@ -253,3 +253,34 @@ func waitForInclusionProof(ctx context.Context, client *http.Client, req *api.Ce
...
+	var cert api.InclusionCert
+	if err := cert.UnmarshalBinary(p.CertificateBytes); err != nil {
+		return fmt.Errorf("failed to decode inclusion cert: %w", err)
+	}
+	key, err := req.StateID.GetTreeKey()
+	if err != nil {
+		return fmt.Errorf("failed to derive SMT key: %w", err)
+	}
+	return cert.Verify(key, req.CertificationData.TransactionHash.DataBytes(), rootRaw, api.InclusionProofV2HashAlgorithm)
+}
```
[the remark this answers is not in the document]
jait91: Fixed the duplication by extracting the SMT-only check into a shared internal helper used by both CLIs

internal/storage/redis/commitment.go
```diff
@@ -814,3 +792,66 @@ func (cs *CommitmentStorage) StreamCertificationRequests(ctx context.Context, co
...
+		if err := pendingRes.Err(); err != nil {
+			if errors.Is(err, redis.Nil) {
+				break
+			}
+			if isNoGroupError(err) {
+				if ensureErr := cs.ensureConsumerGroup(ctx); ensureErr != nil {
+					return fmt.Errorf("failed to ensure consumer group: %w", ensureErr)
+				}
+				continue
+			}
```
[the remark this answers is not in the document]
jait91: XGroupCreate is synchronous; no real tight-loop risk

internal/storage/redis/commitment_integration_test.go
```diff
@@ -976,3 +985,301 @@ func (suite *RedisTestSuite) TestGetByRequestIDs_FindsAcrossMultipleBatches() {
...
+	}
+	require.NoError(t, suite.storage.StoreBatch(ctx, secondBatch))
+
+	suite.storage.ResetPendingSweep()
+
+	stream2Ctx, cancel2 := context.WithCancel(ctx)
+	defer cancel2()
+	stream2Ch := make(chan *models.CertificationRequest, 10)
+	go func() {
+		_ = suite.storage.StreamCertificationRequests(stream2Ctx, stream2Ch)
```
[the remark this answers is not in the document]
jait91: Fixed

internal/signing/certification_request_validator_bft_shard_test.go
```diff
@@ -0,0 +1,159 @@
...
+// shardIDFromHex constructs a ShardID from its trailing-1 end-marker hex form.
+func shardIDFromHex(t *testing.T, h string) types.ShardID {
+	t.Helper()
+	var id types.ShardID
+	require.NoError(t, id.UnmarshalText([]byte(h)))
+	return id
+}
+
+// stateIDWithMSB returns a 32-byte state ID whose first byte holds topBits in
+// its top topBitCount bits (remaining bits zeroed).
```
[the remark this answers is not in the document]
jait91: fixed

internal/service/service.go
jait91: added safeguard, v1 cleanup will be done in separate PR

internal/config/config.go
jait91: Fixed. Config.Validate() now errors if SHARDING_MODE=bft-shard and SHARDING_CHILD_SHARD_ID != 0

internal/bft/client.go
jait91: Bumped bft-go-base version where this issue was fixed, and removed the extra checks

internal/smt/smt.go
jait91: Added one-line clarification that a placeholder leaf with nil value counts as not-found.

internal/service/service.go
```diff
@@ -491,10 +501,9 @@ func (as *AggregatorService) GetInclusionProofV2(ctx context.Context, req *api.G
 		CertificateBytes:   certBytes,
 		UnicityCertificate: types.RawCBOR(block.UnicityCertificate),
 	}
-	if err := proof.Verify(&api.CertificationRequest{
-		StateID:           req.StateID,
-		CertificationData: *record.CertificationData.ToAPI(),
-	}); err != nil {
+
+	selfValue := record.CertificationData.TransactionHash.DataBytes()
+	if err := childCert.Verify(key, selfValue, block.RootHash, api.InclusionProofV2HashAlgorithm); err != nil {
```
MastaP: **Self-verification is weaker than before this commit — please reconsider.**

Before `290d7a2` this block decoded `certBytes` (the marshaled cert that actually ships to the client) and verified it against `proof.UCInputRecordHashRaw()` (the UC that actually ships). The new version verifies the pre-marshal `childCert` against `block.RootHash`, which drops coverage in a few places:

1. **Child mode: composition is no longer self-checked.** When `Sharding.Mode == ShardingModeChild`, `cert` above (line 482-492) is the *composed* cert produced by `api.ComposeInclusionCert(block.ParentFragment, childCert, childRoot)`, and that composed cert is what we marshal and ship. `block.RootHash` is the child root, not UC.IR.h. So `childCert.Verify(key, selfValue, block.RootHash, ...)` asserts only the child-side SMT path against the child root — it never exercises `ComposeInclusionCert`, never touches `ParentFragment`, and never binds to the UC the client will verify against. A bug in parent-fragment composition, a stale fragment stored on the block, or a mismatch between the composed cert and UC.IR.h will ship undetected and fail at the client.

2. **CBOR round-trip is no longer exercised.** The previous check did `selfCert.UnmarshalBinary(certBytes)` + `proof.UCInputRecordHashRaw()`, so it caught marshaling bugs and UC-extraction bugs. The new path verifies the in-memory `childCert` struct and the pre-hex `block.RootHash` field, so a regression in `InclusionCert.MarshalBinary` / `UnmarshalBinary` or `UCInputRecordHashRaw` (e.g. an off-by-one in algorithm-imprint handling) would not be caught here either. This is the exact category of bug a self-verification step exists to trip.

3. **Non-inclusion-visibility regression.** In standalone/bft-shard the check is semantically equivalent (cert == childCert, block.RootHash == UC.IR.h by construction), so everything *happens* to still work — but that invariant is load-bearing and not asserted locally. If someone later changes how `block.RootHash` is populated (or how UC.IR.h is sourced), the self-check silently continues to pass on stale expectations.

### Suggested fix

Restore the prior behavior — verify what we ship, against the UC we ship, using the same `cert.Verify` contract the client will run:

```go
var selfCert api.InclusionCert
if err := selfCert.UnmarshalBinary(certBytes); err != nil {
    return nil, fmt.Errorf("self-check: decode cert: %w", err)
}
selfRoot, err := proof.UCInputRecordHashRaw()
if err != nil {
    return nil, fmt.Errorf("self-check: extract UC.IR.h: %w", err)
}
selfValue := record.CertificationData.TransactionHash.DataBytes()
if err := selfCert.Verify(key, selfValue, selfRoot, api.InclusionProofV2HashAlgorithm); err != nil {
    return nil, fmt.Errorf("generated inclusion proof failed self-verification: %w", err)
}
```

This reintroduces the round-trip, exercises composition in child mode, and binds the check to UC.IR.h — i.e. it verifies the exact bytes+UC the client is about to receive, which is the point of a self-check.

### Test gap

Nothing today would catch the child-mode regression: a unit test that seeds a child block with a deliberately broken `ParentFragment` (or corrupted `UnicityCertificate` IR.h) and asserts `GetInclusionProofV2` returns a `self-verification` error would pin the contract and prevent a future "optimization" from dropping it again. Worth adding alongside the fix.

jait91: Removed the self-check entirely since it isn't actually required on the server side and only adds latency

MastaP (approved):

PR #153, opened 2026-05-14
Request validation hardening

### LINKED ISSUE #150: Request validation hardening
Ensure that all inputs to the state identifier computation (in particular the current owner predicate) use canonical CBOR encoding and do not contain any extra data. Allowing any deviations can result in double-spending attacks.

See changes made in https://github.com/unicitynetwork/state-transition-sdk-js/tree/hardening for reference.

jait91:
Closes #150


pkg/api/certification_request_canonical.go
```diff
@@ -0,0 +1,238 @@
...
+	if out.Version != 1 {
+		return fmt.Errorf("unsupported CertificationRequest version: %d", out.Version)
+	}
+	if out.CertificationData.Version != 1 {
+		return fmt.Errorf("unsupported CertificationData version: %d", out.CertificationData.Version)
+	}
+	return nil
+}
+
+const maxCanonicalCBORDepth = 64
```
MastaP: the code below is canonical CBOR handling, should be extracted somewhere

pkg/api/certification_request_canonical.go
```diff
@@ -0,0 +1,238 @@
...
+// ErrCertificationRequestNotCanonical is returned by
+// UnmarshalCanonicalCertificationRequestCBOR when the input is not encoded in
+// canonical (Core Deterministic) CBOR form. Non-canonical encodings are
+// rejected at the public API boundary so the same logical request cannot be
+// expressed by multiple distinct byte sequences.
+var ErrCertificationRequestNotCanonical = errors.New("certification request CBOR is not canonical")
+
+// UnmarshalCanonicalCertificationRequestCBOR decodes data into out and rejects
+// the input if it is not in canonical Core Deterministic CBOR form.
+func UnmarshalCanonicalCertificationRequestCBOR(data []byte, out *CertificationRequest) error {
```
MastaP: move this to `pkg/api/certification_request.go`. func name looks like an overkill. maybe just `UnmarshalCertificationRequestCBOR` (comment explains the expectation to be canonical)

pkg/api/certification_request_canonical.go
```diff
@@ -0,0 +1,238 @@
...
+	if out.CertificationData.Version != 1 {
+		return fmt.Errorf("unsupported CertificationData version: %d", out.CertificationData.Version)
+	}
+	return nil
+}
+
+const maxCanonicalCBORDepth = 64
+
+const (
+	cborMajorUnsignedInt byte = 0
```
MastaP: I haven't checked but shouldn't these consts and functions come from some existing cbor lib? or maybe we need to extract one and share it between bft and aggregator..

MastaP (changes requested):

pkg/api/certification_request_canonical.go
jait91: Extracted the generic canonical CBOR validation into pkg/cbor, so it is no longer certification-request specific

pkg/api/certification_request_canonical.go
jait91: Moved the request-specific wrapper to certification_request.go and renamed it to `UnmarshalCertificationRequestCBOR`

pkg/api/certification_request_canonical.go
jait91: The CBOR library has equivalent internal constants/parsing logic, but they are not exported for us to reuse directly. I moved the constants and validation code into a generic pkg/cbor package instead of keeping them in certification-request code, SDK handles this the same way

MastaP (approved):

PR #156, opened 2026-06-03
Add RocksDB-backed disk SMT for BFT shard mode

### LINKED ISSUE #152: Disk-backed SMT
Port the Rust aggregator’s disk-backed SMT design from rugregator to the Go aggregator to reduce SMT memory growth

jait91:
## Summary

Adds an optional RocksDB-backed SMT backend for single-active BFT shard mode while keeping the existing in-memory SMT as the default.

## What changed

  - Added a backend abstraction so SMT can run in memory or on RocksDB
  - Wired RocksDB SMT into BFT-shard finalization, startup verification, replay, and recovery.
  - Added proof metadata cache / optional precomputed proof storage.
  - Added RocksDB build/config support and guards that reject unsupported HA and child/parent modes for now. HA support will be added in a separate PR.

## Notes

Mongo still stores finalized block history; RocksDB is the local SMT state and is verified against Mongo on startup. This keeps the data needed for future HA catch-up/rebuild work.


internal/service/service.go
```diff
@@ -227,64 +226,126 @@ func (as *AggregatorService) GetInclusionProofV2(ctx context.Context, req *api.G
...
-	if smtInstance == nil {
+	completed := false
+	as.recordProofPath(ctx, proofPathRequest)
+	defer func() {
+		if !completed {
+			as.recordProofPath(ctx, proofPathError)
+		}
+	}()
+
+	unlock, ok := as.roundManager.TryFinalizationReadLock()
```
MastaP: **🔴 High (blocking) — proofs for already-finalized states can return "not ready"; a latent availability race that shows up as a flaky test.**

This replaces the blocking `FinalizationReadLock()` with the non-blocking `TryFinalizationReadLock()`. When finalization holds `finalizationMu`, the read returns an empty proof (`CertificationData == nil`, line ~250). But `FinalizeBlock` writes the `AggregatorRecord` to Mongo **inside** that same critical section. So a client that polls on record existence and then reads the proof can observe the record as finalized while the lock is *still held*, and gets a spurious "not ready" for a fully-provable state. The old blocking lock waited and returned the real proof.

Observable via `internal/service/disk_bft_proof_rocksdb_test.go:89` (`require.NotNil(... CertificationData)`):
```
CGO_CFLAGS=-I$(brew --prefix rocksdb)/include CGO_LDFLAGS="-L$(brew --prefix rocksdb)/lib -lrocksdb" \
  go test -tags rocksdb -run TestDiskSMTBFTShardProofBeforeAndAfterFinalization ./internal/service/
```

> **Update / correction (was previously stated as "fails 5/5 deterministically"):** this reproduces **5/5 locally** (macOS, librocksdb 11.x) but **passes on the GitHub `ubuntu-latest` runner** (librocksdb 8.9.1) — confirmed by the new `test-rocksdb` CI job, which is green. So it is **not** a hard failure everywhere; whether the single read at line 86 lands inside the finalization write-lock window is **timing-dependent** (hardware / RocksDB version / I/O latency). Treat it as a **flaky test masking a real availability race**, not a deterministic break. The race itself is real and matters under production load (longer/more frequent finalization windows ⇒ more "record finalized but proof not yet readable" responses), and a test that is red locally yet green in CI is itself worth fixing.

**Fix options:** (a) keep a short bounded *blocking* wait before falling back to "not ready", so an already-finalized state isn't reported not-ready merely because a later round holds the lock; or (b) move record-store / precompute out of the `finalizationMu` window so "record durably finalized" implies "proof readable"; or (c) if the transient is intended, make the test poll line 86 instead of asserting once — but that weakens the proof-availability guarantee and is worth an explicit decision.

internal/smt/disk/tree.go
```diff
@@ -0,0 +1,738 @@
...
+		})
+		workers := runtime.GOMAXPROCS(0)
+		if workers < 1 {
+			workers = 1
+		}
+		sem := make(chan struct{}, workers)
+		root, err := batchInsert(t.root, items, 0, len(items), 0, ctx, sem)
+		if err != nil {
+			result.Rejected = append(result.Rejected, RejectedLeaf{
+				Index:  -1,
```
MastaP: **🔴 High (blocking) — swallowed `batchInsert` error + `Index:-1` ⇒ panic in the shared caller, and a phantom-accept.**

On a `batchInsert` error (e.g. `ErrStubOnPath`, malformed branch, `NewInternal` failure) `ApplyLeaves` appends `RejectedLeaf{Index: -1}` and then **returns `(result, nil)`** (line ~188) — the error is swallowed. The shared consumer `internal/round/leaf_add.go:64` then does `commitments[rejected.Index]` → `commitments[-1]` → **runtime panic**, crashing the round-processing goroutine (both live-add and precollector paths).

The memory backend never emits a negative index (its rejected indexes come from a real loop index), so this is a **disk-vs-memory parity defect**.

Compounding it: `AcceptedIndexes` is populated *before* the insert (line ~162) and on error `t.root` is left unchanged, so the failed leaves are simultaneously reported *accepted* while `CandidateRoot` is the **old** root — a silent consistency/data-loss hazard if the panic were ever guarded away.

**Fix:** return `(ApplyResult{}, err)` on `batchInsert` failure (and don't add to `AcceptedIndexes` until after success); add an `Index >= 0` guard in `leaf_add.go`. See also the related comment on `disk_backend.go` (`fromDiskApplyResult`).

Dockerfile
```diff
@@ -17,13 +66,33 @@ RUN go mod download
...
-RUN make build
+RUN if [ -n "$BUILD_TAGS" ]; then \
+      go build -tags "$BUILD_TAGS" -o /app/bin/aggregator ./cmd/aggregator; \
+    else \
+      go build -o /app/bin/aggregator ./cmd/aggregator; \
+    fi
+
+FROM builder AS test
+
+RUN if [ -n "$BUILD_TAGS" ]; then \
```
MastaP, line 77: **🔴 High — the entire disk-SMT test suite never runs in CI, which is how the two bugs above slipped through.**

This `FROM builder AS test` stage runs `go test -tags rocksdb`, but it's an **orphan**: the final image stage `COPY --from=builder` (not `from=test`), no compose file or `docker/build-push-action` sets `--target test`, so BuildKit never builds it. Separately, `.github/workflows/ci.yml` and `docker.yml` run bare `make test` / `make build`, and the `Makefile` defaults `GO_BUILD_TAGS` empty — so every `//go:build rocksdb` file (the whole disk store, backend, snapshot/commit/fork, root parity, recovery/replay, **and the failing `internal/service` test**) is excluded. `make vet` is hardcoded `go vet ./...` too, so the gated files aren't even vetted → silent bit-rot risk.

Even if this stage *did* run, it only covers `./internal/smt/disk/...` and would still miss the `internal/service` / `internal/round` gated tests.

**Fix:** add a CI job that installs librocksdb and runs `make test GO_BUILD_TAGS=rocksdb` (ideally with `-race`), mirroring the `rocksdb-builder` stage. At minimum `go build -tags rocksdb ./...` + `go vet -tags rocksdb ./...` to stop bit-rot.

internal/smt/disk/rocksstore/store.go
```diff
@@ -0,0 +1,1113 @@
...
+	if s.nodesCF != nil {
+		C.rocksdb_column_family_handle_destroy(s.nodesCF)
+		s.nodesCF = nil
+	}
+	if s.metaCF != nil {
+		C.rocksdb_column_family_handle_destroy(s.metaCF)
+		s.metaCF = nil
+	}
+	if s.db != nil {
+		C.rocksdb_close(s.db)
```
MastaP, line 360: **🟠 Medium — `Store.Close()` frees RocksDB handles without synchronization (TOCTOU use-after-free).**

`Close()` sets `s.db = nil` / frees the CF handles and `readOpts` **without holding `s.mu`**, while the read paths touch the same fields unlocked: `getCFWithReadOptions` (nil-check ~693, `rocksdb_get_cf(s.db,...)` ~714), `NewReadSnapshot` (`rocksdb_create_snapshot(s.db)` ~408), and `ReadSnapshot.Close` (~436). The `s.db != nil` checks are a TOCTOU window: an in-flight `GetInclusionProofV2` can pass the check, then `Close()` frees `s.db` before the `C.rocksdb_get_cf` call → use-after-free of a freed `rocksdb_t*`.

Normal shutdown is safe by ordering (HTTP server stops before `roundManager.Stop`), so the trigger is narrow — an `http.Server.Shutdown` **timeout** with a slow proof read still running. But the field accesses are genuinely unsynchronized.

**Fix:** guard the db/CF lifetime — take `s.mu` in `Close()` (and `RLock` in the read paths), or use a `closed` atomic + `WaitGroup`/refcount so `Close` blocks until outstanding reads/`ReadSnapshot`s are released.

internal/smt/backend/disk_backend.go
```diff
@@ -0,0 +1,368 @@
...
+		}
+	}
+	return out
+}
+
+func fromDiskApplyResult(result disk.ApplyResult) BatchApplyResult {
+	rejected := make([]RejectedLeaf, len(result.Rejected))
+	for i, item := range result.Rejected {
+		rejected[i] = RejectedLeaf{
+			Index:  item.Index,
```
MastaP, line 526: **🟠 Medium — `fromDiskApplyResult` leaks the disk backend's `-1` sentinel index through the abstraction boundary.**

This copies `item.Index` verbatim with no validation, so the `Index:-1` produced by `disk/tree.go`'s internal-error path (see the `tree.go` comment) escapes to callers that assume every `Rejected.Index` is a valid slice index — which the memory backend always guarantees. This is the abstraction-layer half of the same parity defect.

**Fix:** treat `item.Index < 0` here as a whole-batch internal error returned via the error channel of `AddLeavesClassified`, so disk and memory backends stay behaviorally identical at the interface (and the shared `leaf_add.go` consumer can't index `commitments[-1]`).

bft-sharding-compose.yml
```diff
@@ -197,6 +203,7 @@ services:
     volumes:
       - ./data/bft-sharding/genesis:/app/bft-config
       - ./logs/bft-shard0:/app/logs
+      - ./data/bft-sharding/smt-rocksdb-shard0:/app/data/smt-rocksdb
```
MastaP, line 206: **🟠 Medium — RocksDB data dir not pre-created/chmod'd ⇒ startup EACCES when `SMT_BACKEND=rocksdb`.**

This bind-mounts `./data/bft-sharding/smt-rocksdb-shard0` (and `-shard1`, line ~309) into a container that runs as a **non-root** user (`${USER_UID:-1001}`), and `SMT_DISK_PATH` points at the mount itself. But the `docker-run-bft-sh-clean*` Makefile targets only `mkdir`/`chmod 777` the genesis/mongo/redis dirs — never these `smt-rocksdb-*` dirs. Since they don't exist when `chmod` runs, Docker auto-creates the bind-mount sources at `up`-time as **root:root 0755**, and `rocksstore.Open()` (non-root) fails with `EACCES`, crashing the aggregator at startup. The default `SMT_BACKEND=memory` hides it, so it only bites the documented rocksdb opt-in.

**Fix:** add `./data/bft-sharding/smt-rocksdb-shard0 ./data/bft-sharding/smt-rocksdb-shard1` to the `mkdir -p` + `chmod -R 777` lines in both bft-sharding make targets.

MastaP (commented):
## Automated-assisted deep review (disk SMT)

I reviewed this with a multi-agent pass (per-subsystem + adversarial verification) **and built/ran the suites locally, including the `-tags rocksdb` tests that CI does not run.**

**Good news first:** all 7 prior bot findings (FIFO cache, the various snapshot-leak sites, `block_syncer` discard) are genuinely fixed; the backend abstraction and disk tree are clean; startup verification is fail-closed.

**However**, two blocking issues are currently masked by the fact that **no disk-SMT test runs in CI**:

- 🔴 The PR's own `TestDiskSMTBFTShardProofBeforeAndAfterFinalization` fails **5/5 deterministically** (see comment on `service.go`).
- 🔴 A disk-vs-memory parity defect can panic the round goroutine (see comment on `disk/tree.go`).

Test results I observed: default build/vet/tests ✅; `-tags rocksdb` for `smt/disk*`, `smt/backend`, `internal/round` ✅ (backend + rocksstore also clean under `-race`); `internal/service` ❌.

The 6 inline comments below cover findings ordered by severity. The in-memory default path is unaffected, so risk is scoped to `SMT_BACKEND=rocksdb` deployments — i.e. the feature this PR ships.

internal/service/service.go
jait91: Fixed by removing the non-blocking proof-read fallback for disk SMT. Disk proof reads now use a published RocksDB snapshot for the last finalized root, so reads stay consistent and do not return false “not ready” while a new block is finalizing

internal/smt/disk/tree.go
jait91: Fixed. Internal disk apply failures now propagate as errors instead of being reported as rejected leaves, and negative indexes are rejected defensively before callers can use them.

Dockerfile
jait91: Fixed. CI now exercises the RocksDB build/test path, and we added RocksDB-tagged vet as an extra guard. The one vet issue it exposed in MultiGet pointer handling was fixed.

internal/smt/disk/rocksstore/store.go
jait91: Fixed. RocksDB store close is now synchronized with all CGO handle users. `Close()` cannot free RocksDB handles while reads, writes, batches, or read snapshots are still using them, and a test covers the long-lived snapshot case.

internal/smt/backend/disk_backend.go
jait91: Fixed. `fromDiskApplyResult` now treats negative rejected indexes as backend errors instead of passing them through the abstraction boundary, so callers only ever see valid rejected indexes.

bft-sharding-compose.yml
jait91: Fixed. Local clean targets now create the RocksDB bind-mount directories and clean them together with Mongo/Redis state, so RocksDB cannot accidentally retain stale SMT data across clean runs.

internal/service/service.go
```diff
@@ -227,65 +226,124 @@ func (as *AggregatorService) GetInclusionProofV2(ctx context.Context, req *api.G
...
-	rootHashRaw := api.HexBytes(smtInstance.GetRootHashRaw())
-	block, err := as.storage.BlockStorage().GetLatestByRootHash(ctx, rootHashRaw)
-	if err != nil {
-		return nil, fmt.Errorf("failed to get latest block by root hash: %w", err)
+	var childCert *api.InclusionCert
+	var rootHash []byte
+	if usesPublishedProofView {
+		rootHash, childCert, err = publishedProofReader.GetPublishedInclusionCert(ctx, key)
+		if err != nil {
+			return nil, fmt.Errorf("failed to build inclusion cert for state ID %s: %w", req.StateID, err)
```
MastaP, line 313: **🔴 High (blocking for the rocksdb path) — disk-mode proof reads hard-error for not-yet-included states, where memory returns a graceful empty proof.**

The published-view path builds the inclusion cert **eagerly**, before the record-existence check:

```go
rootHash, childCert, err = publishedProofReader.GetPublishedInclusionCert(ctx, key)
if err != nil {
    return nil, fmt.Errorf("failed to build inclusion cert for state ID %s: %w", req.StateID, err) // ← hard error
}
```

`persist.BuildInclusionCert` returns a hard error for any absent key (`disk SMT persist: leaf not found`, or `published proof view is empty` for an empty root) — non-inclusion proofs aren't implemented. The **memory** path avoids this by deferring `GetInclusionCert` until *after* it confirms `record != nil && record.BlockNumber <= block.Index`, then returns `emptyInclusionProofResponse` (the documented "Non-inclusion is not implemented yet" contract → verifier short-circuits `ErrExclusionNotImpl`).

So in `SMT_BACKEND=rocksdb`, `GetInclusionProofV2` returns a 500-class error instead of a graceful empty response whenever the state isn't yet in the published tree:

- **Submit → round-pickup window (the normal client poll pattern):** `markProofsPending` only runs when a commitment is *added to a round snapshot* (`batch_processor.go:59`), not at submit time. A client that submits then immediately polls gets a hard error until the next round picks the state up.
- **Never-submitted / unknown / wrong-shard states:** deterministic hard error.
- **Before the first finalized block:** empty published view → hard error.
- **`[markProofsReady, Publish]` finalization window:** brief hard error for the just-finalized state (`batch_processor.go` `FinalizeBlock`: `markProofsReady` runs before `preparedProofView.Publish`).

This is the same *class* of proof-availability symptom as the original `TryFinalizationReadLock` finding, just relocated to the cert-build step — and broader, because it's deterministic for unknown states and covers the entire submit→pickup window, not only the lock-held window.

**Verified locally against a non-empty published tree** (both layers):
- Backend: `DiskBackend.GetPublishedInclusionCert(absentKey)` → `disk SMT persist: leaf not found`.
- Service: `GetInclusionProofV2` for a never-submitted state → `failed to build inclusion cert for state ID …` (hard error), where the memory backend returns an empty proof.

**Why the gated test doesn't catch it:** `TestDiskSMTBFTShardProofBeforeAndAfterFinalization` waits via `require.Eventually` until `GetKnownNotReadyBlock(req2)` is `true` *before* asserting the empty-proof behavior — so it only exercises the graceful known-not-ready early-return, never the eager `GetPublishedInclusionCert` path for an absent key. The "after" assertion is `Eventually`-polled too, which swallows the transient window errors.

**Suggested fix:** mirror the memory path's ordering — resolve root + block + record first, and build the inclusion cert only after `record != nil && record.BlockNumber <= block.Index`, all against the **same** pinned `diskProofView` (capture it once under `proofViewMu.RLock` so root and cert stay consistent). Equivalently, have `GetPublishedInclusionCert` translate "leaf not found" into a non-inclusion signal `(root, nil, nil)` and let the existing `record == nil || record.BlockNumber > block.Index` check produce the empty response — taking care **not** to fall through to the live `GetInclusionCert` at line ~333 (it reads a possibly newer committed root and would re-introduce a root/cert mismatch). A regression test asserting an empty (not error) response for a not-yet-included state in disk mode would lock this down.

---
*Note: all six findings from the prior review round are correctly addressed, and the rest of the disk-SMT design is clean — this is the one new regression introduced by the published-view redesign. Full `-tags rocksdb` suite (incl. `-race` on `rocksstore`/`backend`) passes locally; CI `test-rocksdb` is green.*

jait91: This is fixed in #160, which targets this branch as its base. The disk proof path now returns the same graceful empty not-ready proof as memory mode for absent/not-yet-included leaves instead of hard-erroring

cmd/aggregator/main.go
```diff
@@ -356,16 +357,37 @@ func startLeaderChangedEventListener(ctx context.Context, log *logger.Logger, cf
...
+				if err != nil {
+					log.WithComponent("ha-listener").Error("Failed to verify leadership before activation", "error", err)
+					resignLeadership(ctx, log, leader)
+					continue
+				}
+				if !isStillLeader {
+					log.WithComponent("ha-listener").Warn("Leadership was lost before activation")
+					continue
+				}
 				if err := roundManager.Activate(ctx); err != nil {
```
MastaP, line 388: **🔴 High — a dropped LeaderChanged event can leave a lock-holding node permanently deactivated (shard halt).**

EventBus subscriber channels are cap-1 and `Publish` drops on a full channel; `markLeader`/`markFollower` dedupe via `isLeader.Swap`, so events are edge-triggered and never re-published. This listener blocks for long stretches (promotion sync, `VerifyLeadership`, and `roundManager.Activate` here — recovery + BFT bootstrap can take tens of seconds) and never re-reads actual leadership state.

Sequence: (1) listener blocks in `Activate`; (2) transient `UpdateHeartbeat` error → `markFollower` publishes `false` (buffered in the cap-1 channel); the lock is NOT released, so after TTL the polling loop re-acquires → `markLeader` publishes `true` → channel full → **dropped**; (3) `Activate` returns, listener processes the stale `false` → `Deactivate`. End state: this node renews the leadership lock forever while deactivated — no node produces blocks until manual restart.

The cap-1/drop bus is pre-existing, but this PR makes the listener's blocking window much longer and adds new transition sources. Fix direction: after handling each event (or periodically), reconcile against `leader.IsLeader()`/`VerifyLeadership` instead of trusting the edge stream — i.e. make the listener level-triggered.

MastaP (commented):
## Re-review after the #160 merge (HA enablement)

Verified all prior threads at HEAD (`a86e3b2`) — **every previously reported issue is genuinely fixed**, including the June-10 published-view finding: the path is now `PublishedRoot` + `GetPublishedInclusionCertAtRoot(expectedRoot, key)` with typed sentinels mapped to the graceful empty response, cert built under `proofViewMu.RLock` against the pinned view, no fallthrough to a live root, and the new tests pin all four failure scenarios. The six round-2 fixes (FIFO cache recordRef, all snapshot-discard sites, `ApplyLeaves` error propagation + negative-index guards, `rocksstore` Close synchronization, compose dirs) all survived the merge intact, and the Makefile also covers the new HA compose bind-mount dirs.

**However**, the ~3.2k-line HA merge itself (ccf84d5..a86e3b2) introduces new issues — the 5 inline comments below, ordered by severity. The two 🔴 ones are liveness regressions in leadership transitions, which is exactly the surface this merge enables; none of them have test coverage (they all live in transition windows the HA test stack doesn't exercise).

Also worth noting (not inline):
- `round_manager.go` `Activate` HA recovery commits the recovered block via `LoadRecoveredNodesIntoBackend` but never calls `refreshDiskProofView` (unlike `Start()`), so a promoted leader serves a stale published view for ~1 round.
- `block_syncer.go` `updateSMTForBlock`: if `Commit` succeeds but `RefreshPublishedProofView` fails, the next tick skips the already-committed block and never refreshes — published view stays stale indefinitely on a quiet chain. Consider refreshing whenever committed root ≠ published root.
- `service.go` `GetInclusionProofV2`: `PublishedRoot` and `GetPublishedInclusionCertAtRoot` are two separate lock acquisitions; a view rotation between them yields a once-per-round spurious "not ready" for a long-included state. A single retry of the pair would eliminate it.
- `persist/tree.go:532,575`: nil-subtree / unmaterialized-stub traversal outcomes now classify as `ErrProofLeafNotFound`, so an internal invariant break would degrade silently to "not included" instead of a loud error — consider logging the stub case.
- `TestDiskSMTHAFollowerCatchesUpAndPromotes` races `follower.Activate`'s real 100ms stub round against the manual finalize of round 2 — flake risk on slow runners.

Test status: default suite and `-tags rocksdb` gated suite (smt/round/service/ha) pass locally when run in isolation; vet clean both tagged and untagged; CI `test-rocksdb` green. (Running both suites concurrently on one Docker daemon flakes on testcontainers — local-only artifact.)

internal/round/round_manager.go
```diff
@@ -425,68 +555,90 @@ func (rm *RoundManager) StartNewRoundWithSnapshot(
...
-			"previousRoundAge", time.Since(rm.currentRound.StartTime).String())
+			"previousRoundNumber", currentRoundNumberString,
+			"previousRoundState", currentRoundState.String(),
+			"previousRoundAge", currentRoundAge.String())
+		if currentRoundNumber != nil && currentRoundNumber.Cmp(roundNumber.Int) >= 0 {
+			rm.roundMutex.Unlock()
+			if snapshot != nil {
+				snapshot.Discard(ctx)
+			}
+			rm.logger.WithContext(ctx).Info("Skipping stale round start",
```
MastaP, line 609: **🔴 High — this `>= 0` skip (new in ccf84d5) silently removes the repeat-UC re-propose path; a lost certification request now stalls the shard.**

On a repeat UC (root chain timed out waiting for our certification — e.g. the cert request was lost in transit), `bft/client.go` clears `proposedBlock` and calls `StartNewRound(N)` with `tr.Round == N == currentRound.Number`. That call now hits this skip and returns nil — nothing ever re-collects and re-sends the certification request for round N. Every subsequent repeat UC no-ops the same way, so the shard stalls until failover/restart. The old behavior rebuilt round N and re-sent the request; that was the only retransmission mechanism.

Compounding: `StartNewRound` runs `discardActivePrecollector` *before* this skip decision, so each repeat UC also destroys the next-round precollector and resets the Redis sweep for nothing.

The skip is correct for the cross-round HA double-start case (`>`), but the equal-round repeat-UC case needs to re-propose — e.g. allow `==` when the current round is in proposed state with no UC, or give the BFT client an explicit re-propose entry point. No test covers the repeat-UC equal-round path.

internal/bft/client.go
```diff
@@ -346,8 +332,11 @@ func (c *BFTClientImpl) handleUnicityCertificate(ctx context.Context, uc *types.
 		return nil
 	}
 
-	// Check for repeat UC (same InputRecord but higher root round)
-	if c.isRepeatUC(prevLUC, uc) {
+	isRepeat, err := uc.IsRepeat(prevLUC)
```
MastaP, line 338: **🟠 Medium — `handleUnicityCertificate` finalizes without checking the UC certifies *our* root.**

Before `FinalizeBlockWithRetry`, the only guard is round-number equality (`proposedRound != expectedRound`); `uc.InputRecord.Hash` is never compared with `c.proposedBlock.RootHash`. In the HA overlap window this PR newly enables (old leader not yet deactivated while the new leader activates — both propose round N with different roots X and Y), the loser can match the winner's UC to its own proposal and write a block with `RootHash=Y` but a UC certifying X into shared Mongo. If the loser's write lands first, the winner hits `validateDuplicateBlock` root-mismatch and fatals, leaving the inconsistent block in the chain (followers replay it fine — `verifySMTForBlock` checks only `block.RootHash`, never the UC).

A one-line `bytes.Equal(uc.InputRecord.Hash, c.proposedBlock.RootHash)` guard before finalizing (treat mismatch like a repeat UC: clear proposal, start next round) closes this. The new `ErrStaleCertificationRound` rejection covers the cross-round variant but not this same-round one.

internal/service/service.go
```diff
@@ -495,6 +583,47 @@ func (as *AggregatorService) GetHealthStatus(ctx context.Context) (*api.HealthSt
 	return modelToAPIHealthStatus(status), nil
 }
 
+func (as *AggregatorService) addDiskProofReadiness(ctx context.Context, status *models.HealthStatus) {
```
MastaP, line 586: **🟠 Medium — the readiness gate (f006f44) is fail-open for diverged and far-behind followers.**

`addDiskProofReadiness` only flags `CommittedState().BlockNumber == nil`, which covers just the empty-RocksDB window. Two gaps:

1. **Divergence never surfaces**: a follower hitting `ErrDiskSMTDiverged` fails every sync tick forever, but the error is only a log line in `runLoop` — `BlockNumber` stays non-nil, so the node reports healthy and keeps serving proofs from a permanently frozen view. On the leader path divergence is fatal; on the follower path there's no escalation at all.
2. **Catch-up window**: `syncDiskToBlock` commits per block, so a fresh follower 100k blocks behind reports ready the moment block 1 commits, then serves an extremely old root for the rest of the catch-up.

A lag bound (`latest.Index - committed.BlockNumber <= K`) plus marking unhealthy on persistent sync failure/divergence would close both.

internal/round/round_manager.go
```diff
@@ -938,16 +1208,27 @@ func (rm *RoundManager) Deactivate(ctx context.Context) error {
...
 	}
-	rm.clearProofPending()
 	rm.stopCommitmentPrefetcher()
 	if rm.bftClient != nil {
 		rm.bftClient.Stop()
 	}
+	rm.roundWG.Wait()
+
+	rm.roundMutex.Lock()
+	rm.currentRound = nil
```
MastaP, line 1306: **🟠 Medium — Deactivate racing an in-flight UC finalization escalates a graceful step-down into full process shutdown.**

`bftClient.Stop()` cancels the message-loop context but doesn't join the goroutine, and `roundWG` doesn't cover the UC handler — so `handleUnicityCertificate → FinalizeBlockWithRetry` can still be running when this sets `currentRound = nil`. The new strict preconditions in `FinalizeBlock` ("no active round" / round-number mismatch / snapshot-root mismatch) then fail deterministically *before* the block is stored; `FinalizeBlockWithRetry`'s recovery finds no unfinalized block, exhausts its 3 attempts, and the BFT client publishes `TopicFatalError` — which main.go treats as process shutdown, on a node that just stepped down cleanly. (The ctx-cancel variant pre-exists; the new early validations + `currentRound = nil` make it deterministic in the deactivation window.)

Fix: treat finalize failures after deactivation (`precollectorDisabled` set / ErrDeactivated-class) as non-fatal, or join the message loop in `Stop()` before `Deactivate` proceeds.

MastaP:
**Addendum — one more find while re-running the suites: `TestChildMode_RequiresFreshParentProof` fails reproducibly (3/3) when the round package is built with `-tags rocksdb`, passes untagged. CI cannot see this combination.**

Failure: `precollection_test.go:1441` — block 2 is finalized but `rm.lastAcceptedParentUCRound` reads 1 instead of 2.

Root cause is an assertion race in the test, amplified by the tagged build, not a freshness-logic bug: in `batch_processor.go` the child path runs `FinalizeBlockWithRetry(block)` and only **then** `rm.acceptParentUC(parentUC)`. The test's `require.Eventually` fires as soon as block 2 appears in storage — which happens inside `FinalizeBlockWithRetry`, before `acceptParentUC` — and then immediately asserts the atomic. The window between block-stored and UC-recorded is the finalize tail (SetFinalized + ack), which is a few ms in the untagged build but ~20–40 ms in the cgo/rocksdb build on my machine — wide enough that the 25 ms `Eventually` poll lands in it essentially every run.

Why CI is blind to it: the untagged `test` job is fast enough to win the race, and the `test-rocksdb` job only discovers `//go:build rocksdb`-gated tests — `precollection_test.go` is untagged, so the failing combination (untagged test + tagged binary) is never built in CI. Same class of gap as the original Dockerfile finding, one level down.

Fix is in the test: make the final assertion part of the `Eventually` condition (`block != nil && rm.lastAcceptedParentUCRound.Load() == 2`) or assert on `block.Finalized` rather than mere existence. Optionally, have the CI `test-rocksdb` job run the full `internal/round` package with the tag, not just the gated files — that would have caught this.

(For completeness: `TestChildPreCollection_CommitmentAfterProofBeforeRoundEnd_ShouldBeInNextRound` also failed once under the tagged full-suite run but passes 2/2 in isolation — ordinary load flake, same timing-sensitivity family. Everything else in both suites is green in isolation: untagged 18/18 packages, tagged smt/backend/rocksstore/persist/service/ha all ok.)

cmd/aggregator/main.go
jait91: I’m keeping this PR scoped to disk-SMT correctness fixes; the remaining HA lifecycle/readiness issues will be handled together in a follow-up hardening PR

internal/round/round_manager.go
jait91: Fixed. Equal-round repeat UCs now retry the existing finalizing proposal instead of being skipped, and StartNewRound no longer tears down precollector state unless it is actually abandoning an older round

internal/bft/client.go
jait91: Fixed. The BFT client now checks uc.InputRecord.Hash against proposedBlock.RootHash before finalization and fails closed on mismatch

PR #158, opened 2026-06-05
fix: prevent SMT corruption during HA leadership churn

jait91:
Storage pressure could cause HA leadership churn (the aggregator instance repeatedly lost and reacquired
  leadership) while old child-round goroutines were still waiting for parent proofs. A stale finalizer could then write the current round's pending SMT data under an older block, causing blocks, block_records, aggregator_records, and smt_nodes to diverge.

Fix by cancelling and waiting for in-flight round processing on Deactivate, refusing to overwrite a same-or-newer active round on reactivation, and validating round number, snapshot root, and duplicate-block contents before finalization continues.

### DIFF
--- a/internal/round/finalize_duplicate_test.go
+++ b/internal/round/finalize_duplicate_test.go
@@ -592,11 +592,10 @@ func (s *FinalizeDuplicateTestSuite) Test7_FinalizeBlockRejectsRoundNumberMismat
 	}
 }
 
-// Test8_DuplicateFinalizedBlockMustNotCommitDifferentSnapshot verifies that a
-// retry for an already-finalized block is only idempotent if it is replaying the
-// same block state. A duplicate block must not commit whatever snapshot happens
-// to be in currentRound.
-func (s *FinalizeDuplicateTestSuite) Test8_DuplicateFinalizedBlockMustNotCommitDifferentSnapshot() {
+// Test8_DuplicateFinalizedBlockRejectsSnapshotRootMismatch verifies that a
+// duplicate block retry cannot commit whatever snapshot happens to be in
+// currentRound when that snapshot does not match the block root.
+func (s *FinalizeDuplicateTestSuite) Test8_DuplicateFinalizedBlockRejectsSnapshotRootMismatch() {
 	t := s.T()
 	ctx, cancel := context.WithCancel(context.Background())
 	defer cancel()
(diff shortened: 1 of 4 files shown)


internal/round/round_manager.go
```diff
@@ -416,13 +428,15 @@ func (rm *RoundManager) StartNewRoundWithSnapshot(
...
 
+	rm.roundWG.Add(1)
 	rm.roundMutex.Unlock()
 
 	rm.logger.WithContext(ctx).Info("Started new round",
 		"roundNumber", roundNumber.String(),
 		"commitments", len(commitments))
 
 	rm.wg.Go(func() {
+		defer rm.roundWG.Done()
```
MastaP, line 439: **High — an HA stand-down during the child *finalize* window is misclassified as fatal and crashes the node.**

This PR newly makes `Deactivate` cancel the round's `activeCtx` (line 893). That correctly unblocks `pollForParentProof` when leadership is lost *during the poll*: it returns `context.Canceled`, `processRound` wraps it with `%w`, and the `errors.Is(err, context.Canceled)` guard on line 441 treats it as a clean shutdown. ✅ (This is the path the new test covers.)

It is **not** handled when leadership is lost *during finalization* — parent proof already received, `FinalizeBlockWithRetry` running under the now-cancelled `activeCtx`:

1. `FinalizeBlock`'s guards pass (`currentRound` is still non-nil — `Deactivate` is blocked here in `roundWG.Wait()`), then `storeBlockAndRecords(activeCtx)` fails on the cancelled context.
2. `FinalizeBlockWithRetry` (`batch_processor.go:366-400`) never checks for cancellation: its `GetUnfinalized(activeCtx)` recovery probe also fails on the cancelled ctx (so recovery is skipped), it `time.Sleep`s ignoring ctx, retries 3× (~2s), and finally returns a plain `fmt.Errorf("FinalizeBlock failed after %d attempts")` that does **not** wrap `context.Canceled`.
3. `processRound` returns that error, so the guard on line 441 is **false**, and line 450 publishes `TopicFatalError` → `startFatalErrorListener` (`cmd/aggregator/main.go:320`) cancels the app context and the whole process exits.

Net: a clean follower transition that happens to land in the finalize window stalls ~2s on the ctx-ignoring `time.Sleep`s and then **crashes the node** — and under the storage pressure this PR targets, finalize is slower (wider window), so churn can crash-loop. The PR's own non-fatal-shutdown guard (line 441) is silently defeated by `FinalizeBlockWithRetry` flattening the cancellation into a generic error.

**Suggested fix:** make `FinalizeBlockWithRetry` bail out on cancellation and propagate it — e.g. `if ctx.Err() != nil { return ctx.Err() }` at the top of the loop and after a failed attempt, and replace `time.Sleep(finalizeRetryDelay)` with `select { case <-time.After(finalizeRetryDelay): case <-ctx.Done(): return ctx.Err() }`. Regression test: hold the parent proof, `Deactivate` *after* releasing it so finalize runs under a cancelled ctx, and assert no `TopicFatalError` is published.

internal/round/finalize_duplicate_test.go
```diff
@@ -507,3 +508,160 @@ func (s *FinalizeDuplicateTestSuite) Test6_BlockRecordsMatchPendingCommitmentsOn
...
+		assert.Len(t, rm.currentRound.PendingCommitments, 3)
+		assert.NotNil(t, rm.currentRound.Snapshot)
+	}
+}
+
+// Test8_DuplicateFinalizedBlockMustNotCommitDifferentSnapshot verifies that a
+// retry for an already-finalized block is only idempotent if it is replaying the
+// same block state. A duplicate block must not commit whatever snapshot happens
+// to be in currentRound.
+func (s *FinalizeDuplicateTestSuite) Test8_DuplicateFinalizedBlockMustNotCommitDifferentSnapshot() {
```
MastaP: **Medium — this test exits at the snapshot-root guard, so the new duplicate-mismatch validation has zero coverage.**

Test8 pre-stores block 71 with `RootHash = emptyRoot` but sets `currentRound.Snapshot` to a non-empty pending root (asserted `!= emptyRoot` in the setup). In `FinalizeBlock` the round-number guard passes (71 == 71), then the **snapshot-root guard** (`batch_processor.go:459`) compares the non-empty snapshot root against `block.RootHash` (emptyRoot); they differ, so it returns `finalizeRejectSnapshotRoot` — **before** ever reaching `storeBlockAndRecords` / `validateDuplicateBlock` (`batch_processor.go:494-498`).

So despite its name, Test8 re-tests the snapshot-root guard, and the duplicate-content validation this PR adds is never exercised: `validateDuplicateBlock`'s root-mismatch (line 702), `sameStateIDs == false` (line 714), and the `finalizeRejectDuplicateMismatch` rejection (lines 498-503) are covered by no test — reverting those checks wouldn't fail CI.

**Suggested:** add a test where `block.RootHash` **matches** the pending snapshot root (so the snapshot-root guard passes) but the stored `block_records` StateIDs differ from `currentRound.PendingCommitments`, then assert `FinalizeBlock` returns the duplicate-mismatch error and persists nothing extra. Consider relabeling Test8 to reflect that it currently covers the snapshot-root guard.

MastaP (commented):
Two findings from a focused review of the child-sharding HA path. Context: in child mode the child `RoundManager` has no BFT client (it talks to the parent via the root client), and finalization is driven by `processRound` → `pollForParentProof` → `FinalizeBlockWithRetry`, which `Deactivate`'s new `roundWG.Wait()` correctly covers — and cancel-during-poll is handled and tested. These two are additional to the existing automated-review comments: one High (a leadership stand-down landing in the *finalize* window is misclassified as fatal and crashes the node) and one Medium (the new duplicate-mismatch guard is untested because Test8 exits early at the snapshot-root guard).

MastaP:
@jait91

PR #166, opened 2026-07-03
perf: black-box load/verification client for deployed aggregator/gateway

MastaP:
## What

`cmd/perf-blackbox` — a purely client-side load and inclusion-verification tool for a deployed aggregator or subscription gateway. Unlike `cmd/performance-test`, it scrapes **no** `/metrics` and reads **no** server logs; every number it reports is observed from the client.

## Wire protocol

Speaks exactly what the state-transition SDK speaks:

- `certification_request` — params = `hex(CBOR(CertificationRequest))` (tag 39030, v1)
- `get_inclusion_proof.v2` — params = `{"stateId":"<32-byte hex>"}`
- `X-API-Key` header carries the shared API key
- `X-State-ID` header carries the state-id hex (the gateway routes on it)

## Behaviour

- Generates **uniform**, cryptographically valid commitments (fresh key + random state per request, no shard rejection sampling), so the gateway spreads them evenly across all shards.
- Submits at a target rate, then after a configurable delay fetches each inclusion proof back and verifies it locally against the SMT root (`internal/proofverify.VerifyInclusionProofLocal`) — confirming every accepted commitment actually landed in the tree.
- Reports: submit/accept rates, submit and submit→included latency percentiles, per-shard-prefix distribution (uniformity check), and an error breakdown.

## Flags

`-url -api-key -rate -duration -count -proof-delay -proof-retry -proof-timeout -verify -proofs -submit-concurrency -proof-concurrency -conns -http-timeout -insecure -gen-workers`

`-proofs=false` measures accept rate only (used for rate ramps).

## Build / run

Repo needs Go 1.25+.

```
go build -o bin/perf-blackbox ./cmd/perf-blackbox
./bin/perf-blackbox -url https://<gateway>/ -api-key "$UNICITY_API_KEY" \
    -rate 1000 -duration 10s -proof-delay 1.5s -verify
```

The API key is only ever read from `-api-key` / `UNICITY_API_KEY`; nothing is hardcoded.

Findings from testnet2 runs are in a comment below.

### DIFF
--- a/cmd/perf-blackbox/main.go
+++ b/cmd/perf-blackbox/main.go
@@ -1,0 +1,800 @@
+// Command perf-blackbox is a pure black-box load/verification client for a
+// deployed aggregator (or subscription gateway fronting several shards).
+//
+// It speaks exactly the wire protocol the state-transition SDK speaks:
+//   - certification_request  params = hex(CBOR(CertificationRequest))  (tag 39030, v1)
+//   - get_inclusion_proof.v2 params = {"stateId":"<32-byte hex>"}
+//   - X-API-Key   header carries the shared API key
+//   - X-State-ID  header carries the 32-byte state id hex (the gateway routes on it)
+//
+// It scrapes NO /metrics and reads NO server logs. Everything reported is
+// observed from the client: submission results, and — after a delay — inclusion
+// proofs fetched back and cryptographically verified, to confirm every accepted
+// commitment actually landed in the SMT.
+//
+// Commitments are uniform: each carries a fresh key pair and random state, so
+// the derived state ids are uniformly distributed and the gateway spreads them
+// evenly across all shards. No shard rejection-sampling is done.
+package main
+
+import (
+	"context"
+	"crypto/rand"
+	"crypto/tls"
+	"encoding/json"
+	"flag"
+	"fmt"
+	"log"
+	"net/http"
+	"os"
+	"runtime"
+	"sort"
+	"strings"
+	"sync"
+	"sync/atomic"
+	"time"
+
+	"github.com/btcsuite/btcd/btcec/v2"
+	"golang.org/x/net/http2"
+
+	"github.com/unicitynetwork/aggregator-go/internal/proofverify"
+	"github.com/unicitynetwork/aggregator-go/internal/signing"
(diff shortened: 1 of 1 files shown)


MastaP:
## testnet2 findings

Runs against the 8-shard subscription gateway (`https://gateway.testnet2.unicity.network/`) with a shared test key.

### Correctness ✅

- Smoke (30 @ 30/s) and baseline (3000 @ 300/s): **100% accepted, 100% included in SMT, 100% of proofs cryptographically verified.**
- Load spreads uniformly across all 8 shard prefixes (`000`–`111`).
- Baseline latency (300/s, clean): submit p50 58ms / p95 109ms; submit→included p50 2.52s / p95 3.03s (dominated by the 1.5s proof delay + ~900ms block cadence).

### Throughput ceiling ⚠️

The gateway sheds submissions with `HTTP 503 "Service Overloaded"` far below the key's configured rate limit.

| Offered | Submitted | Accepted | Accept % | Shed (503) |
|---:|---:|---:|---:|---:|
| 300/s  | 3000  | 3000 | 100%  | 0    |
| 350/s  | 4200  | 4200 | 100%  | 0    |
| 400/s  | 4800  | 4480 | 93.3% | 320  |
| 1000/s | 10000 | 3533 | 35.3% | 6467 |

- Knee is between **350 and 400 req/s**; sustained accepted throughput ≈ **350–375/s**.
- Everything accepted is aggregated correctly (100% SMT inclusion at every rate) — the shards are **not** the bottleneck.
- This is ~27× below the shared key's 10,000 req/s limit, so the **API-key rate limiter is not the cause**.

### Interpretation

Each ramp step above is a fresh ~12s burst. A horizontal autoscaler reacts to sustained load over minutes, so a 12s burst likely hits a **single** gateway instance and never triggers scale-out. ~375/s is therefore plausibly single-instance capacity (a per-instance concurrency / in-flight cap emitting `503 Service Overloaded`), with autoscaling never engaging.

### Next step

A **sustained** run (e.g. `-rate 1000 -duration 180s -proofs=false`) while watching the gateway instance count + HAProxy logs will distinguish:

- **autoscaling lag** — accepted/s climbs from ~350 toward ~1400 as instances spin up, vs.
- **a fixed per-instance or global cap** — accepted/s stays pinned at ~350/s.

Issue #135, opened 2026-03-15
[TGE] Unicity ID Predicate support (Aggregator)


martti007:
Besides PayToPublicKeyPredicate support we need predicate to support sending tokens to specific unicity id. This includes the verigification of the unicity id token.

PR #170, opened 2026-07-09
Switch SMT to big-endian bit ordering

### LINKED ISSUE #169: Sync Aggregator SMT bit ordering with Yellowpaper big-endian bit strings
## Summary

The Yellowpaper RSMT/RSMST specification now defines keys, regions, and inclusion bitmaps using big-endian bit-string notation:

- SMT keys are 32-byte values read as 256-bit big-endian bit strings.
- Depth `d` addresses bit `d` of that big-endian bit string.
- `rsmt_sort_key(k) = k`; ordinary unsigned lexicographic byte order is the tree traversal order.
- Inclusion certificate bitmaps are 256-bit big-endian bit vectors.
- Regions are packed as 256-bit big-endian bit strings: the first `d` bits are the key prefix and the suffix is zero.

The Go Aggregator still has explicit LSB-in-byte assumptions in the SMT implementation, proof verification, sharding prefix checks, and disk SMT encoding. Update the implementation and tests to match the spec.

## Why

The change should make Go follow the natural bit order used by ordinary byte comparisons and big-endian bit-string notation.

## Current Hotspots

Likely places that need changes:

- `pkg/api/state_id.go`
  - `PathToFixedBytes` and `FixedBytesToPath` currently reverse byte order for an LSB-first SMT key layout.
  - Comments explicitly say `key bit d is bit (d%8) of key[d/8]`.

- `pkg/api/state_id_bitorder_test.go`
  - Tests pin LSB-first semantics (`TestFixedBytesToPath_UsesLSBFirstBitAddressing`, `RoundtripLSBFirst`).

- `pkg/api/inclusion_cert.go`
  - `verifyBitmapPath` tests bitmap bits using `(bitmap[d/8] >> (d%8)) & 1`.
  - `keyBitAt` uses LSB-in-byte.
  - `RegionFromKeyBytes` packs prefixes by copying low-order bits in each byte and masking with `(1 << rem) - 1`.
  - Comments mention LSB-first layout.

- `pkg/api/inclusion_cert_compose.go`
  - `bitmapDepthRange` uses LSB-in-byte bitmap indexing.

- `pkg/api/smt.go`
  - `RegionFromPathBits` packs `path.Bit(i)` into `region[i/8] |= 1 << (i % 8)`.
  - Comments describe v6a region bits as LSB-in-byte.

- `pkg/api/shard_match.go`
  - `MatchesShardPrefix` checks LSB-first key bits and reads expected shard bits from the low bits of `shardBitmask`.
  - This should be reconciled with the spec's shard-prefix interpretation.

- `internal/smt/smt.go`
  - `populate` sets region bits with `rightRegion[depth/8] |= 1 << (depth % 8)`.
  - Parent-mode shard tree construction depends on this bit convention.

- `internal/smt/disk/hash.go`
  - `HashNode` comment says region is packed LSB-in-byte.
  - `RegionFromKey` and `KeyBit` use LSB-in-byte.

- `internal/smt/disk/path.go`
  - `CompressedPath` construction, `BitAt`, and unused-bit canonicalization use LSB-in-byte packing.

- `internal/smt/disk/node_key.go`
  - `clearUnusedPrefixBits` masks unused bits as if prefix bits are LSB-in-byte.
  - Persisted node keys include prefix bytes, so this affects disk compatibility and migration decisions.

- `internal/smt/disk/persist/tree.go`
  - Inclusion certificate generation sets bitmap bits with `1 << (depth % 8)`.
  - `setPrefixBit` uses LSB-in-byte prefix packing.

- Test/vector files
  - `internal/smt/golden_vectors_test.go`
  - `internal/smt/v6a_interop_vectors_test.go`
  - `internal/smt/yellowpaper_hash_semantics_test.go`
  - `pkg/api/inclusion_cert_test.go`
  - `internal/smt/disk/*_test.go`
  - Any golden roots or certificates generated under the old convention must be regenerated or explicitly marked legacy.

## Implementation Notes

Prefer introducing small shared helpers instead of repeating shifts everywhere:

- `KeyBit(key []byte, d int) byte`: `return (key[d/8] >> (7 - uint(d%8))) & 1`
- `SetBitBE(bits []byte, d int)`: `bits[d/8] |= 0x80 >> uint(d%8)`
- `ClearSuffixBE(prefix []byte, bitLen int)` for region/bitmap/path canonicalization.
- `RegionFromKeyBytes(key []byte, depth int)` should copy the prefix bytes and keep the high `depth % 8` bits in the boundary byte.

For non-byte-aligned prefixes, the old LSB mask `(1 << rem) - 1` becomes a high-bit mask, e.g. `0xff << (8 - rem)` when `rem != 0`.

## Migration / Compatibility

Changing bit order changes:

- SMT root hashes.
- Inclusion certificate bitmap bytes.
- Internal-node region bytes, hence node hashes.
- Shard routing and split behavior.
- Persisted disk SMT node keys and stored roots.
- Existing golden vectors.

No backwards compatibility needed. Old-format disk-backed SMT data can be deleted.

## Acceptance Criteria

- All SMT key-bit access, bitmap access, region packing, and prefix packing in Aggregator use big-endian bit-string semantics.
- `FixedBytesToPath` / `PathToFixedBytes` round-trip without reversing the 32-byte key order relative to spec key bytes.
- Inclusion certificate verification and generation use big-endian bitmap bits.
- RSMT node hashes match the updated Yellowpaper region packing.
- Batch/traversal ordering is ordinary unsigned lexicographic byte order (`rsmt_sort_key(k)=k`).
- Shard-prefix matching and shard split/parent proof logic are aligned with the same bit convention.
- Disk SMT metadata/versioning prevents old-layout persisted data from being silently interpreted as new-layout data.
- Tests cover boundary depths: `0`, `1`, `7`, `8`, `200`, `255`.
- Golden vectors are regenerated for the new convention, with any retained old vectors clearly labeled legacy.

## Out of Scope

- Changing CBOR serialization structure or public proof envelope shapes.
- Changing hash domain separators.
- Changing key byte order on the wire; only bit indexing within the existing byte string changes.

jait91:
Syncs the SMT with the yellowpaper "big-endian bit strings" convention: key
bit d is the MSB-first bit (key[d/8] >> (7-d%8)) & 1, regions and inclusion-
cert bitmaps are big-endian bit vectors, and rsmt_sort_key(k) = k — so disk
traversal order is plain lexicographic byte order (keyPathLess collapses to
bytes.Compare). Wire byte order, hash domain separators, node/leaf hash
layouts, and CBOR/JSON-RPC shapes are unchanged.

All bit math routes through three helpers in pkg/api (KeyBitBE, SetBitBE,
ClearSuffixBE). The in-memory big.Int path machinery is untouched — only
FixedBytesToPath/PathToFixedBytes change, from a byte-order reversal to a full
256-bit reversal (bits.Reverse8 + byte swap), so path bit i still equals the
routing decision at depth i. Leaf hashes and the empty/one-leaf roots stay
byte-identical, but every internal node hash, root, cert bitmap and tree shape
changes.

No back-compat: TreeLayout bumps to yellowpaper-rsmt-sha256-v6a-be — pre-flip
data refuses to open, so all trees (incl. perf snapshots) need a rebuild from
finalized history, and shard routing (key→shard) must be re-derived at
rollout.

Vectors (internal/smt/v6a_interop_vectors_test.go): empty, one-leaf, shallow
split, deep split (diverge at bit 255), multi-leaf, and 6-leaf parity —
regenerated under big-endian with asymmetric canaries. The pkg/api helper
tests pin boundary depths 0/1/7/8/200/255, and an independent big-endian
reference (be_reference_anchor_test.go, raw SHA-256) anchors production to the
spec. Cross-verified both ways against the JS SDK
(unicitynetwork/state-transition-sdk-js#135): identical root c854db11… for the
shared 6-leaf parity tree, and their BitString uses the same MSB-first
convention. Java and Rust must reproduce at least the deep-split and
multi-leaf roots (multi-byte region packing) before the coordinated rollover.

Closes #169

### DIFF
--- a/internal/smt/disk/hash.go
+++ b/internal/smt/disk/hash.go
@@ -49,7 +49,7 @@ func HashLeaf(key Key, value []byte) Hash {
 // HashNode implements the yellowpaper v6a (Appendix C.3.2.2) internal-node
 // hash: H(0x01 || depth_1B || region_32B || left_hash_32B || right_hash_32B).
 // The region is the node's absolute key prefix at its bifurcation depth,
-// packed LSB-in-byte with all bits at positions >= depth cleared.
+// packed big-endian with all bits at positions >= depth cleared.
 func HashNode(left, right Hash, depth uint8, region PrefixBits) Hash {
 	h := sha256.New()
 	_, _ = h.Write([]byte{0x01, depth})
@@ -63,15 +63,14 @@ func HashNode(left, right Hash, depth uint8, region PrefixBits) Hash {
 
 // RegionFromKey packs the depth-bit prefix of a descendant key into the
 // canonical v6a region encoding: key bits 0..depth-1 in place, all bits at
-// positions >= depth cleared.
+// positions >= depth cleared. Under big-endian bit ordering the region shares
+// the key's byte layout, so it is the key bytes with the depth suffix masked.
 func RegionFromKey(key Key, depth uint8) PrefixBits {
 	var region PrefixBits
 	d := int(depth)
 	byteLen := (d + 7) / 8
 	copy(region[:byteLen], key[:byteLen])
-	if rem := d % 8; rem != 0 {
-		region[byteLen-1] &= byte(1<<uint(rem)) - 1
-	}
+	api.ClearSuffixBE(region[:], d)
 	return region
 }
 
@@ -82,10 +81,12 @@ func EmptyRootHash() Hash {
 	return Hash{}
 }
 
-// KeyBit returns bit d of key using the yellowpaper/Go v2 LSB-first key layout.
+// KeyBit returns bit d of key using the yellowpaper big-endian bit layout:
+// bit 0 is the MSB of byte 0, bit d = (key[d/8] >> (7 - d%8)) & 1. This is the
+// disk-typed accessor mirroring api.KeyBitBE.
 func KeyBit(key Key, d int) byte {
 	if d < 0 || d >= KeyBits {
 		panic(fmt.Sprintf("disk smt: key bit index out of range: %d", d))
 	}
-	return (key[d/8] >> (uint(d) % 8)) & 1
+	return (key[d/8] >> (7 - uint(d)%8)) & 1
 }
--- a/internal/smt/disk/hash.go
+++ b/internal/smt/disk/hash.go
@@ -83,10 +83,10 @@ func EmptyRootHash() Hash {
 
 // KeyBit returns bit d of key using the yellowpaper big-endian bit layout:
 // bit 0 is the MSB of byte 0, bit d = (key[d/8] >> (7 - d%8)) & 1. This is the
-// disk-typed accessor mirroring api.KeyBitBE.
+// bounds-checked disk-typed accessor over api.KeyBitBE.
 func KeyBit(key Key, d int) byte {
 	if d < 0 || d >= KeyBits {
 		panic(fmt.Sprintf("disk smt: key bit index out of range: %d", d))
 	}
-	return (key[d/8] >> (7 - uint(d)%8)) & 1
+	return api.KeyBitBE(key[:], d)
 }
--- a/internal/smt/disk/persist/tree_test.go
+++ b/internal/smt/disk/persist/tree_test.go
@@ -3,6 +3,8 @@
 package persist
 
 import (
+	"encoding/binary"
+	"math/rand"
 	"testing"
 
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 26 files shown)


MastaP (commented):
Two findings from a review of the big-endian switch. The conversion itself looks uniform and internally consistent across the in-memory, disk, and cert-verify paths (independent BE reference + JS-SDK parity vectors both anchor it), and the buildable suites pass. These are the two items worth acting on before merge.

internal/smt/disk/persist/tree_test.go
```diff
@@ -322,10 +322,12 @@ func TestSnapshotCommitPersistsNonRootLoadedNodeMovement(t *testing.T) {
...
 	tree := openPersistTree(t, store)
 
-	k0 := keyWithFirstByte(0x00) // root-left leaf
-	k1 := keyWithFirstByte(0x07) // root-right subtree, bits 1110...
-	k2 := keyWithFirstByte(0x0f) // root-right subtree, bits 1111...
-	k3 := keyWithFirstByte(0x01) // root-right subtree, splits non-root path at bit 1
+	// Big-endian first bytes chosen so the tree shape matches the original
+	// LSB-first fixture (bytes are the bit-reversal of 0x00/0x07/0x0f/0x01).
+	k0 := keyWithFirstByte(0x00) // root-left leaf (big-endian bit 0 = 0)
+	k1 := keyWithFirstByte(0xE0) // root-right subtree, big-endian bits 111000...
```
MastaP, line 330: **Test coverage gap: the persist proof path is never exercised past byte 0.**

Every key in this package's tests comes from `keyWithFirstByte(...)`, so all splits happen within byte 0 (depth < 8). But the persisted path has its *own* `keyPathLess` (rewritten to `bytes.Compare`) and bitmap packing (`api.SetBitBE(cert.Bitmap[:], depth)` in `buildInclusionCert`), and this is the path that actually serves inclusion proofs from RocksDB. A big-endian indexing bug that only manifests at depth ≥ 8 — wrong bitmap byte index, or a `bytes.Compare` mis-ordering beyond byte 0 — would pass this entire suite yet emit invalid proofs for realistic hash-distributed keys (splits routinely land past byte 0). `disk.Tree` guards exactly this with `TestTreeApplyRandomizedParityWithMemory` (400 random full-width keys); the persist package has no equivalent, and since it's `//go:build rocksdb` it isn't covered by the default `go test` run either.

Suggest adding a randomized full-width parity test here mirroring `disk.Tree`'s: random 32-byte keys → `CandidateRoot == memoryRootAfterLeaves`, plus `GetInclusionCert(key).Verify(...)` for every key across a reopen. That single test closes the cross-byte gap for ordering, bitmap packing, and materialization at once.

internal/smt/disk/hash.go
```diff
@@ -82,10 +81,12 @@ func EmptyRootHash() Hash {
...
-// KeyBit returns bit d of key using the yellowpaper/Go v2 LSB-first key layout.
+// KeyBit returns bit d of key using the yellowpaper big-endian bit layout:
+// bit 0 is the MSB of byte 0, bit d = (key[d/8] >> (7 - d%8)) & 1. This is the
+// disk-typed accessor mirroring api.KeyBitBE.
 func KeyBit(key Key, d int) byte {
 	if d < 0 || d >= KeyBits {
 		panic(fmt.Sprintf("disk smt: key bit index out of range: %d", d))
 	}
-	return (key[d/8] >> (uint(d) % 8)) & 1
+	return (key[d/8] >> (7 - uint(d)%8)) & 1
```
MastaP: **Nit: `KeyBit` open-codes `api.KeyBitBE` instead of delegating.**

This line is byte-for-byte the body of `api.KeyBitBE`, and the comment even says it "mirrors api.KeyBitBE" — yet the sibling accessor `CompressedPath.BitAt` in `path.go` was changed to delegate (`return api.KeyBitBE(p.bits[:], pos)`). Since `disk.Key` is `[32]byte`, this could do the same after the bounds check: `return api.KeyBitBE(key[:], d)`. `bits.go` states its goal is "a single audit point governs the whole bit convention"; leaving a second hand-maintained copy of the MSB-first formula is a drift risk. Not a bug — the two are identical today.

PR #173, opened 2026-07-28
Add prefix-major disk SMT layout

jait91:
## Summary

- Add an opt-in prefix-major RocksDB node layout with persisted layout validation.
- Replace the startup replay block limit with paginated, resumable replay of all missing finalized history.

Prefix-major ordering improves locality for path-oriented SMT access. Large-tree benchmarks showed lower materialization and finalization latency than depth-major while preserving identical roots and proofs.

**Migration:** Enabling prefix-major requires a new or empty `SMT_DISK_PATH` rebuilt from finalized history. Existing depth-major RocksDB data cannot be reused, but MongoDB and BFT state do not require a reset.


MastaP (commented):
Reviewed the prefix-major layout itself carefully and it looks sound: the encoding is injective (canonical zero-fill via `clearUnusedPrefixBits` guarantees `(prefix, depth)` uniqueness), the MultiGet index-permutation reorder in `getNodesWithReadOptionsOpen` is correct, forcing `sortedInput = true` is safe because there's no custom RocksDB comparator or prefix extractor (bytewise order == `bytes.Compare`), and the nodes CF is point-read only so no traversal can depend on key ordering. Persisted layout tag + `depth-major` default everywhere means existing databases are unaffected. The lifecycle parity test is good coverage.

My concerns are all in the replay change rather than the headline feature.

### 1. Unbounded startup replay is an undocumented behavior change

Dropping `SMT_STARTUP_REPLAY_LIMIT_BLOCKS` means a node that is far behind now attempts the entire replay — one Mongo round trip per 100 blocks plus one `CommittedState` read and one commit *per block* — where it previously failed fast with "explicit rebuild required". That may well be the right call, but it turns a seconds-long failure into a potentially hours-long startup, and neither the PR description nor the README mentions it. At minimum document it; consider a wall-clock budget as the replacement guard rail.

### 2. Progress logging can go silent for the whole replay

`disk_smt_startup.go:293`

```go
if replayed%1000 == 0 || cursor.Cmp(latestBlock.Int) == 0 {
```

This is only evaluated at page boundaries. With gaps (repeat-UC rounds — which `TestDiskSMTStartupPaginatedReplaySkipsMissingRepeatUCRounds` establishes are normal) pages return fewer than 100 blocks, so `replayed` may never land on a multiple of 1000. Combined with #1, an hours-long replay can emit exactly two log lines. Suggest a `nextLogAt` threshold that advances past whatever value is actually hit, or just log once per page.

### 3. Cancellation during replay is recorded as a startup failure

There's no `ctx.Done()` check in the replay loop. A SIGTERM mid-replay fails `GetFinalizedPage`, which routes through `diskSMTStartupFailure` and increments `SMTStartupRecoveryActions{action="fail"}`. That's a spurious failure metric on every graceful shutdown, and #1 widens the window from seconds to hours. An explicit cancellation branch at the top of the loop returning `ctx.Err()` without bumping the counter would fix it.

### 4. Scope / description

The title and body describe only the layout change, but the diff also removes `ROUND_DURATION`, adds `BLOCK_SYNC_INTERVAL`, changes the `BlockStorage` interface (`GetRange` → `GetFinalizedPage`), and removes the replay cap. The migration note covers only RocksDB. Worth either splitting or expanding the description — a reviewer going by the body alone won't know to look at #1. Also worth noting in release notes that `ROUND_DURATION` is now silently ignored for anyone who had tuned it (the default is unchanged at 1s, so only tuned deployments are affected).

Separately, the benchmark claim ("lower materialization and finalization latency") has no numbers in the description and no benchmark in the tree. Not blocking while the flag defaults off, but those figures should be recorded somewhere before anyone proposes flipping the default.

jait91:
@MastaP Thanks, addressed the startup replay concerns. The README now documents that replay is unbounded, paginated, and resumable; progress logging uses a threshold that cannot be skipped; and graceful cancellation exits without recording a startup failure. Added a regression test proving cancellation preserves committed progress and resumes successfully.

MastaP:
Checked `05574d4` — all three items are addressed.

1. **Unbounded replay** is now documented in the README (no block-count limit, may take significant time, per-block commits make it resumable) and the PR body has a summary bullet for it. Good enough for me; a wall-clock budget would still be nice eventually but I only asked for the docs.
2. **Progress logging** — the `nextLogAt` threshold plus catch-up loop is correct: it fires on the first page that crosses each 1000-block boundary regardless of page size, so gaps can no longer swallow every progress line.
3. **Cancellation** — the `ctx.Err()` reclassification keeps the `action="fail"` counter clean, and `TestDiskSMTStartupPaginatedReplayCancellationResumesWithoutFailureMetric` pins it properly (asserts `context.Canceled`, asserts the counter is unchanged, then asserts replay resumes from block 101). No `t.Parallel()` in `internal/round`, so the global-counter delta check won't be flaky.

Three notes on the fix commit, none blocking:

- **`go.mod` is no longer tidy.** The new test imports `github.com/prometheus/client_model` directly but it's still listed `// indirect`; `go mod tidy` moves it into the direct block. CI only runs `go vet -tags rocksdb` and `go test -tags rocksdb`, so nothing will catch this — worth a `go mod tidy` before merge.
- **The `ctx.Err()` checks are dense** — nine of them, four firing after *successful* operations. Only the error-path ones do real work (reclassifying a storage failure as cancellation); the loop-top check already covers the rest. Functionally harmless, just noisy.
- **SIGTERM during replay still exits 1.** `main.go:199-201` treats any `roundManager.Start` error as fatal-with-code-1, so shutdown mid-replay still logs `Failed to start round manager` and exits non-zero. The failure metric — the thing I actually flagged — is fixed; a truly clean shutdown would need main.go to special-case `context.Canceled`. Pre-existing for every startup path, so fine to leave.

Two minor items from my read that I didn't raise the first time, take or leave:

- `loadOrInitMetadata` returns `&LayoutMismatchError{Got: ""}` when the `treeLayout` metadata key is absent, so a truncated or partially initialized store reports "unsupported disk SMT tree layout" and answers `true` to `IsLayoutMismatch`. An operator reading that will go change `SMT_NODE_KEY_FORMAT` when the real problem is a damaged database. Worth splitting the `!ok` case into its own error.
- In `TestDiskBackendRocksDBNodeKeyFormatLifecycleParity`, `results[format]` is only assigned if the subtest completes, so a `require` failure inside `runDiskBackendRocksDBLifecycle` leaves the parent dereferencing a zero-valued struct at `depth.finalState.BlockNumber.String()` — nil panic that buries the real assertion. A `require.Len(t, results, 2)` before the comparisons would keep failures readable.

PR #174, opened 2026-07-28
Fix round recovery and MongoDB transaction retries

jait91:
## Summary

  - Harden BFT client startup, shutdown, HA deactivation, and durable-proposal recovery.
  - Clear superseded round state while preserving unrelated proof-pending markers.
  - Reset Redis pending sweeps for unresolved rounds and prevent resets from being lost during an active sweep.
  - Detect wrapped MongoDB transaction labels and correctly retry transient transaction and commit failures.

  Includes regression, race, storage integration, and BFT lifecycle coverage.

### DIFF
--- a/internal/bft/client.go
+++ b/internal/bft/client.go
@@ -412,40 +412,51 @@ func (c *BFTClientImpl) handleUnicityCertificate(ctx context.Context, uc *types.
 
 	wasInitializing := c.status.Load() == initializing
 	if wasInitializing {
-		c.logger.WithContext(ctx).Info("BFT client initialization finished, starting first round",
+		c.logger.WithContext(ctx).Info("BFT client initialization finished",
 			"nextRoundNumber", nextRoundNumber.String())
 		// First UC received after an initial handshake with a root node -> initialization finished.
 		c.status.Store(normal)
-		recovered, err := c.finalizeCertifiedDurableProposalLocked(ctx, uc)
-		if err != nil {
-			return err
-		}
-		if recovered {
-			c.logger.WithContext(ctx).Info("Durable proposal finalized from initialization UC",
-				"ucRound", uc.GetRoundNumber(),
+
+		// If startup already proposed a block and this UC certifies it, continue
+		// through the normal proposed-block finalization path below. Otherwise
+		// the initialization branch can skip finalization and start the next
+		// round with a live proposed block still in memory.
+		if c.proposedBlock != nil && c.proposedBlock.Index.Uint64() == uc.GetRoundNumber() {
+			c.logger.WithContext(ctx).Info("Initialization UC matches proposed block; finalizing normally",
+				"blockNumber", c.proposedBlock.Index.String(),
 				"nextRoundNumber", nextRoundNumber.String())
-			err := c.roundManager.StartNewRound(ctx, api.NewBigInt(nextRoundNumber))
+		} else {
+			recovered, err := c.finalizeCertifiedDurableProposalLocked(ctx, uc)
 			if err != nil {
-				c.logger.WithContext(ctx).Error("Failed to start first round after durable proposal recovery",
+				return err
+			}
+			if recovered {
+				c.logger.WithContext(ctx).Info("Durable proposal finalized from initialization UC",
+					"ucRound", uc.GetRoundNumber(),
+					"nextRoundNumber", nextRoundNumber.String())
+				err := c.roundManager.StartNewRound(ctx, api.NewBigInt(nextRoundNumber))
+				if err != nil {
+					c.logger.WithContext(ctx).Error("Failed to start first round after durable proposal recovery",
+						"nextRoundNumber", nextRoundNumber.String(),
+						"error", err.Error())
+				}
+				return err
+			}
+			resumed, err := c.resumeDurableProposalLocked(ctx, api.NewBigInt(nextRoundNumber))
+			if err != nil {
+				return err
+			}
+			if resumed {
+				return nil
+			}
+			err = c.roundManager.StartNewRound(ctx, api.NewBigInt(nextRoundNumber))
+			if err != nil {
+				c.logger.WithContext(ctx).Error("Failed to start first round after initialization",
 					"nextRoundNumber", nextRoundNumber.String(),
 					"error", err.Error())
 			}
 			return err
 		}
-		resumed, err := c.resumeDurableProposalLocked(ctx, api.NewBigInt(nextRoundNumber))
-		if err != nil {
-			return err
-		}
-		if resumed {
-			return nil
-		}
-		err = c.roundManager.StartNewRound(ctx, api.NewBigInt(nextRoundNumber))
-		if err != nil {
-			c.logger.WithContext(ctx).Error("Failed to start first round after initialization",
-				"nextRoundNumber", nextRoundNumber.String(),
-				"error", err.Error())
-		}
-		return err
 	}
 
 	// Check if we have a proposed block that matches the UC round
--- a/internal/bft/client.go
+++ b/internal/bft/client.go
@@ -30,6 +30,7 @@ const (
 	idle status = iota
 	initializing
 	normal
+	stopping
 )
 
 var (
@@ -37,6 +38,7 @@ var (
 	ErrCertifiedRootMismatch   = errors.New("certified root does not match proposed block root")
 	ErrInvalidUCSequence       = errors.New("invalid unicity certificate sequence")
 	ErrCertifiedStateMismatch  = errors.New("local committed state does not match latest certified root")
+	ErrBFTClientStopping       = errors.New("BFT client is stopping")
 )
 
 // BFTClientImpl handles communication with the BFT root chain via P2P network
@@ -49,7 +51,7 @@ type (
 		logger      *logger.Logger
 		eventBus    *events.EventBus
 
-		// mutex for peer, network, signer TODO: there are readers without mutex
+		// Protects peer, network, and signer.
 		mu      sync.Mutex
 		peer    *network.Peer
 		network *BftNetwork
@@ -70,6 +72,8 @@ type (
 		ucProcessingMutex sync.Mutex
 
 		msgLoopCancelFn context.CancelFunc
+		msgLoopDone     chan struct{}
+		stopDone        chan struct{}
 
 		// timestamp when last UC was received
 		lastCertResponseTime atomic.Int64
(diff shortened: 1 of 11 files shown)


MastaP:
## Review notes

Verified against the branch (`go vet` clean, `internal/bft` + the new Mongo label test pass locally; the two new `internal/round` tests couldn't run in my env — the testcontainer times out on index creation and the pre-existing `TestStartNewRoundAbandonsSupersededPendingRound` fails the same way, so it's environmental).

The Mongo fix looks solid. Two concerns on the round-recovery side, the first blocking.

### 1. The BFT init fix looks broken in the only state that produces it

The new branch (`internal/bft/client.go:424`) falls through to the normal proposed-block path, which for a non-resumed proposal ends at `FinalizeBlockWithRetry` (`client.go:652`). That requires an active round matching the block — `batch_processor.go:565-575`: `"cannot finalize block %s: no active round"`.

How can `status == initializing` and `c.proposedBlock != nil` coexist?

- `initializing` is only entered by `Start()` (`client.go:168`), called only from `RoundManager.Activate` (`round_manager.go:1276`).
- Non-HA activates once at boot (`main.go:231`) with `proposedBlock == nil`. So the state needs an HA leader flap: `Deactivate` → `Activate` (`main.go:388/393`).
- `Stop()` (`client.go:210-234`) clears peer/network/signer but **not** `proposedBlock` / `resumedDurableProposal` — that's the actual leak.
- `Deactivate` sets `rm.currentRound = nil` (`round_manager.go:1394`), and `Activate` for standalone/bft-shard starts only the BFT client and prefetcher — no round is started before the first UC.

So when the new branch fires: `rm.currentRound == nil`, and `resumedDurableProposal == false` (normal proposals go through `CertificationRequest`, which sets it false). `FinalizeBlockWithRetry` then fails "no active round"; `GetUnfinalized` is empty because HA `Activate` already rolled the proposal forward to finalized (`recoverUnfinalizedBlock` → `setBlockFinalizedWithCertificate`, `recovery.go:340`); so it burns 3 attempts × 1s and returns an error → `TopicFatalError` → **process shutdown** (`main.go:306-336`), with no round started.

The pre-fix code survived this: `finalizeCertifiedDurableProposalLocked` saw `localRoot == uc.InputRecord.Hash`, returned `false`, and started the next round. `TestBFTClientInitializationFinalizesMatchingInMemoryProposal` passes only because `stubRoundManager.FinalizeBlockWithRetry` unconditionally succeeds (`client_stub_test.go:48-50`) and can't model the active-round precondition.

Suggested fix: set `c.resumedDurableProposal = true` before falling through, so the path uses `FinalizeCertifiedProposal` — it needs no active round and returns `(true, nil)` when the block is already finalized (`recovery.go:34-50`).

Related: the `else` branch never touches `c.proposedBlock`, so when the first UC after re-activation is for a *different* round (a standby took over and the chain moved on), the leak described in the new comment still happens. Clearing `proposedBlock`/`resumedDurableProposal` in `Stop()` — or abandoning unconditionally at the top of the initializing branch — covers every sub-case including the one above.

Also worth calling out in the PR body: the matching-round-but-root-mismatch case now returns `ErrCertifiedRootMismatch` + fatal event with **no round started** (per `TestBFTClientInitializationRejectsMatchingRoundRootMismatch`). Fine if that's the intended fail-fast, but it's a behavior change.

### 2. `resetPendingSweep` signal is racy

`round_manager.go:796` uses `Snapshot != nil` as a proxy for "this round wasn't finalized" (true because `clearFinalizedRound` nils both, `batch_processor.go:1094-1097`). But `processRound`'s defer also nils `Snapshot` when a round is cancelled without proposing (`round_manager.go:822-826`). A round that collected commitments but flushed none into `PendingCommitments` before cancellation then evaluates to `false`, and its unacked PEL entries are never replayed — the exact leak this fix targets.

`round.Block` is set only by `clearFinalizedRound`, so it's a precise signal:

```go
resetPendingSweep := rm.currentRound.Block == nil &&
    (len(rm.currentRound.PendingCommitments) > 0 || len(rm.currentRound.Commitments) > 0)
```

Two minor notes on the same hunk:

- `resetRedisPendingSweep` vs `requestRedisPendingReplay` (`:1434` / `:1448`) are near-identical names with materially different behavior — the new one skips the prefetcher stop and `drainCommitmentStream()`. I assume that's because it runs under `roundMutex` (safe: `ResetPendingSweep` is just an atomic store, `redis/commitment.go:108`, whereas `stopCommitmentPrefetcherAndWait` would not be). Worth a comment on why the drain is unnecessary here, or move the reset below `rm.roundMutex.Unlock()` next to `supersededSnapshot.Discard(ctx)` and reuse the full helper.
- `clearProofPending()` wipes the whole map, including the incoming round's markers that the precollector already set via the `rm.markProofsPending` callback (`:1518`) before handoff. Reachable when `FinalizeBlockWithRetry` succeeds via its recovery branch (`batch_processor.go:483-489`) without `clearFinalizedRound`. Impact is limited to losing the known-not-ready shortcut (`service.go:254`) — requests fall back to a truthful non-inclusion proof — and `reconcileRecoveredFinalization` already does the same wholesale wipe, so low priority.

### 3. Mongo change — correct, one note

`errors.As` on the `mongo.LabeledError` interface is the right fix: `WriteException`/`BulkWriteException` (how in-transaction write conflicts actually surface) now match, and `%w` wrapping no longer defeats detection. Dropping the `if !ok { return }` early return is safe — non-labeled commit errors fall through to the final `return`. I checked all three `WithTransaction` callers (`recovery.go:208`, `batch_processor.go:1068`, `parent_round_manager.go:643`); they're pure in-transaction writes, so the higher retry rate can't double-apply.

The `!isMongoMaxTimeMSExpiredError(err)` guard is an undocumented fourth change. It matches driver/spec semantics (don't retry commit on MaxTimeMSExpired), so keep it — but please mention it in the PR description and consider a test. Note it only inspects `mongo.CommandError`, so a `WriteException` carrying a MaxTimeMSExpired write-concern error won't match (fine for commit errors in practice).

MastaP:
## Re-review of a8ec8a3 — all three points addressed, at the root

Both concerns from my earlier comment are fixed, and fixed at the cause rather than the symptom. I ran a multi-agent adversarial pass over this revision (13 agents across four lenses: BFT lifecycle/concurrency, round-manager invariants, storage layer, test adequacy) — **nothing survived verification**. Details below in case the refuted claims are useful context.

### 1. BFT init path — resolved

The matching-in-memory-proposal branch is gone and initialization is back to durable-proposal-first, so nothing routes a stale in-memory block into `FinalizeBlockWithRetry` any more. The actual root cause I pointed at — `Stop()` leaving `proposedBlock` behind — is now fixed directly (`client.go:279-283`), which also covers the second half of my comment (the non-matching case, where the stale proposal used to leak silently).

`TestBFTClientResumedDurableProposalFinalizesWithoutActiveRound` pins the property that mattered: `require.Zero(t, rm.finalizeBlockCallCnt)`, i.e. the durable path never touches the active-round-requiring API.

### 2. `resetPendingSweep` signal — resolved

`unresolved := currentRound.Block == nil` is exactly the precise signal, and populating `RecoveryResult.Block` → `currentRound.Block` in `reconcileRecoveredFinalization` closes the recovery-path gap I flagged (`FinalizeBlockWithRetry` succeeding via its recovery branch without `clearFinalizedRound`).

This one is empirically pinned: a verifier agent mutated `round_manager.go:807` back to `Snapshot != nil` and `TestStartNewRoundWithSnapshotAbandonAfterSnapshotCleanupResetsRedisPendingSweep` failed at `require.Equal(t, 1, resetQueue.resetCount)`. The test discriminates the fix rather than just passing alongside it.

Targeted `clearProofsPending` / `clearProofStateIDsPending` also resolve the wholesale-wipe note, with `require.True(t, pending)` on the incoming round's marker.

### 3. Mongo — resolved

`TestIsMongoMaxTimeMSExpiredErrorMatchesWrappedCommandError` covers the guard. Deleting the unreachable `return nil` after the commit loop is a good catch on its own: had it ever been reached it would have reported success without a commit.

### Beyond what I raised

Worth calling out, since they aren't in the PR description:

- **A real lost-reset race in Redis**: a `ResetPendingSweep()` arriving mid-drain was clobbered by the drain's trailing `pendingExhausted.Store(true)`. The generation counter fixes it, and `TestPendingExhausted_ResetDuringActiveSweepIsNotLost` reproduces it properly (I ran it — passes).
- `sendCertificationRequest` now snapshots `peer`/`network`/`signer` under `c.mu`, retiring the "readers without mutex" TODO. I grepped: there are no unsynchronized readers left.
- `Start()` closes a partially initialized peer and returns to `idle`; `BootstrapConnect` moved ahead of publishing the fields.
- `publishFatalError` suppressing `context.Canceled` — without it a graceful HA deactivation mid-finalize would shut the whole process down.
- `Stop()` is now a real drain (`stopping` state, cancel-then-wait on `ucProcessingMutex`, join the msg loop, concurrent-safe via `stopDone`).

### What I verified

- `go vet ./...` and `go build ./...` clean.
- `internal/bft` full suite **with `-race`**: 30/30 pass, including 10 new lifecycle tests.
- `internal/round`: passes. (In a single full-package run `TestParentRoundManagerSuite` failed on a testcontainer i/o timeout — 40+ Mongo containers churning in one run on my box; it passes in isolation and that file is untouched by this revision. My earlier "can't run these" note was also just a too-short container timeout, not the tests.)
- `internal/storage/redis` (incl. the new integration test), `internal/storage/mongodb`, `internal/service`, `pkg/...`: pass.
- Lock ordering: `ucProcessingMutex → c.mu` is consistent between `Stop()` and the UC path, and `Start()` takes only `c.mu`; `roundMutex → proofCacheMu` with no inversion anywhere in `internal/round`.
- Every `rm.currentRound` dereference is nil-guarded, so the new `currentRound = nil` is safe.
- `luc` is `atomic.Pointer`, so `Store(prevLUC)` with a nil `prevLUC` in `rollbackInitialization` is fine (this would have panicked on `atomic.Value`).
- No unbounded path in the now-blocking `Stop()`: `FinalizeBlockWithRetry` re-checks ctx each attempt, `waitBeforePrecollectorHandoff` selects on ctx/done, `AdvanceRound` is guarded by `doneCh`.

### Minor, non-blocking

1. **`Stop()` is now an unbounded blocking wait** on `ucProcessingMutex` + the msg loop, on the HA failover path. I couldn't find a path that fails to observe cancellation, but a bounded wait with a warning log would keep failover robust against a future blocking call sneaking into UC processing.
2. **`abandonSupersededRoundLocked` now nils `rm.currentRound`** — a broader lifecycle change than the reported bug. It checks out (abandon only fires for `Number < roundNumber`, so the `==`/`>` duplicate/stale guards are untouched, and it actually *tightens* `proposeBlock`'s superseded check so a stale proposal can no longer be sent during the `StartNewRound` window). Worth a line in the PR body so the invariant change is on the record.
3. **PR description is now well behind the change.** It still lists the original three fixes; this revision adds BFT lifecycle hardening, the Redis sweep-generation race, the `currentRound = nil` invariant, and the `MaxTimeMSExpired` guard from the previous revision. Worth refreshing before merge.
4. Pre-existing, noted while tracing: when a repeat UC skips a round, the durable proposal for the skipped round keeps `status: "proposed"` and is never abandoned (the abandon is a no-op once `proposedBlock` is nil). It's invisible to `GetByNumber`/`GetLatest`/`GetUnfinalized`, so it's storage garbage rather than a correctness problem — but it accumulates. Same in `main`; separate change if you care.

### Refuted claims, for the record

Recording these because the reasoning is reusable, not because action is needed:

- **"Repeat-UC path skips durable recovery → duplicate-key → crash loop."** Refuted from bft-core: a timeout/repeat `CertificationResponse` always carries `TR.Round = prev + 1` (`sharding.go` `nextRound` increments unconditionally when `req == nil`), so the round holding the stale proposal is *skipped*, never re-proposed — no index collision. `TestBFTClientRepeatUCCompletesInitialization` already encodes that convention (IR round 10, `TechnicalRecord{Round: 12}`).
- **"`StartNewRound` can restart the prefetcher while `Deactivate` stops it."** The plumbing weakness is real (`startCommitmentPrefetcher` never clears `prefetchCancel` when its goroutine exits) but every line involved is byte-identical to `main`, and this revision *narrows* the trigger rather than widening it. Pre-existing; separate change if you want it.
- **"Abandoning an empty in-collect round strands a commitment in the PEL."** The window is a sub-microsecond lock race (both the append and the `len(Commitments)` read are under `roundMutex`), and `trimStream` pins `minID = pending.Lower` so the entry is never trimmed; the next `discardActivePrecollector` replays it regardless of commitment counts.

Looks good to merge from my side once the PR body is refreshed.

